PR #1, opened 2023-02-21
Glides Events and testing / CI

skyqrose:
This PR includes:
- JSON Schema for events emitted by Glides, as proposed in https://github.com/mbta/technology-docs/pull/18
- Examples copied from that RFC, which are validated as test cases
- README
- A test script that validates the schema and the examples using [ajv](https://ajv.js.org/)
- asdf, npm, and prettier
- A GitHub CI workflow

Future JSON Schemas that are added should be able to follow the same format as the Glides schema and examples without having to make any modifications to CI.

The automated tests have already paid off: They caught several errors in the glides schema and examples, which I'll copy over to the RFC.


README.md
```diff
@@ -0,0 +1,38 @@
+# MBTA Schemas
+
+A collection of schemas for inter-app interchange formats. Schemas in this repository should stay up to date if they change.
+
+Schemas go in the `schemas/` folder.
+
+Examples go in the `examples/<schema>/` folders.
```
boringcactus, line 7: What if the schema was in `<schema>/schema.json` and the examples were the other files in `<schema>/`? I think storing the examples alongside the schema might be easier to maintain than having one folder for schemas and one folder for examples, but I'm not sure.

boringcactus (approved):

README.md
skyqrose: Thinking through this...

I guess some use cases might want multiple schema files to reference each other, so having an app folder instead of a single `schemas/<schema>.json` file is good.

I'd probably want `schemas/<app>/<schema>.json` and `schemas/<app>/examples/`, so apps don't get mixed up with all the CI stuff that ended up top level, and so you can tell schema json files apart from example json files.

If there are multiple schemas in a folder, I'm not sure how the tests should know which schema to apply against which example. Perhaps `schemas/<app>/examples/<schema>/<whateverexample>.json`? matches to `schemas/<app>/<schema>.json`?

But that's all overengineered / YAGNI. The important part for now is that the current glides schema is tested, and the example folder isn't a problem at this scale, so I want to leave it as is. If a future use case decides that they need a more featured / more organized naming scheme, then they can do it then. It'd just be file moves, so wouldn't be a big change in git.

paulswartz (commented):
In order to work with the documentation tools:
- Each schema should be its own file, and in a directory named after the event 
- Each example should be its own file 

It also makes it easier for other tooling, since they can find the schema for an event given the `type` without needing to look at each schema file.

skyqrose:
So that's 3 separate files, one for each event, plus a common file for the shared types.

The official docs suggest [bundling](https://json-schema.org/understanding-json-schema/structuring.html#bundling) multiple schemas into a single file. In that case, it'd be one file with 3 subschemas with `$id`s.

For code organization reasons, I think it'd be nice to have a single Glides schema file.

This [non-authoritative stack overflow answer](https://stackoverflow.com/a/65575832) points out that the folder path doesn't really matter, all that matters is the $id.

It also doesn't help that in JSON Schema the $id doesn't have to resolve to the schema. The $id URI is just a name.

But all that's in theory, what matters is how the tools work. Do you have any specific tools in mind so I can look up the docs and confirm what they'll need in order to locate a schema?

skyqrose:
We're starting to think about publishing events again, and so I want to unstick this.

Paul, what were the documentation tools you were thinking of, so I can check that they'll work with whatever organization we go with?

Melody, there have been a few changes since your last review, so I'd like you to take another look.

skyqrose:
(sorry for the review request notifications, apparently if you re-request a review from one person, it removes the request to review from the other? I meant to request you both.)

paulswartz:
They're linked in the Asana ticket about documentation, but I'll include them here as well for @boringcactus: 
- Rendered: https://paulswartz.github.io/schemas/events/com.mbta.ocs.raw_message/
- Source: https://github.com/paulswartz/schemas

Having all the events in a single file means that it's much harder for automated tools to handle the events. Instead of being able to take a given event name and go directly to the file with the schema for that event, it would need to parse all of the JSON Schema files, to find the one where there's a `type` field which matches the given event.

boringcactus:
What Asana ticket are you talking about? Also, when were you planning on submitting that fork as a PR?

skyqrose:
I think he means [this Asana ticket about defining the JSON Schema for OCS events](https://app.asana.com/0/881264583703207/1201653786809384/f)

I looked into EventCatalog, and one major hurdle for using it for Glides is that it doesn't support `$ref`s to other schema files. [The docs](https://www.eventcatalog.dev/docs/events/adding-schemas) don't mention it, and there are a couple open issues about it ([Issue 1](https://github.com/boyney123/eventcatalog/issues/220)) ([Issue 2](https://github.com/boyney123/eventcatalog/issues/197)). This PR would need to have refs to shared schemas because the events share types, and because the cloudevents schema doesn't belong in the Glides events schemas.

I looked into alternatives to EventCatalog, and the best I found was [Docusaurus](https://docusaurus.io/) with a [json-schema plugin](https://jy95.github.io/docusaurus-json-schema-plugin/) whose [configuration docs](https://jy95.github.io/docusaurus-json-schema-plugin/docs/api/JSONSchemaViewer#configuration) mention [custom resolvers](https://github.com/stoplightio/json-ref-resolver).

Some options I see:
- Continue this PR as is without a static docs generator. If we add one later, we can restructure the schemas then.
- Add EventCatalog to this PR, and one of:
  - choose some non-standard way to reuse schemas. The schemas would exist, but not be visible in EventCatalog.
  - duplicate all the shared types and schemas into each event schema.
  - find some yet-unknown workaround, or submit contributions to EventCatalog to get this functionality.
- Use some other docs generator, either Docusaurus or something else.

paulswartz:
I'll take a look at https://docusaurus.io/; thanks for pointing it out!

I think that while the `$ref` makes it easier to DRY-up the schemas, it might make it more complicated for consumers to use. For example, you can fetch `https://paulswartz.github.io/schemas/schemas/<message type>/schema.jsonschema` to get everything you need. But I'll see what happens with the alternative tool; maybe there's a way we can have the build step generate those message type-specific schemas while allowing you to have a single file in the repo.

skyqrose:
Ah, I hadn't considered JSON Schema bundling in a build step. That's [definitely possible](https://json-schema.org/blog/posts/bundling-json-schema-compound-documents), and seems like it could solve most of our problems.

Let me know when you've investigated Docusaurus and/or bundling and have a recommendation on what to do for this PR.

paulswartz:
Does this PR block anything on the Glides side?

skyqrose:
No. It takes up a bit of time and attention to stay on top of an open PR, so I'd like to not let it sit indefinitely, but it's not blocking anything.

paulswartz:
In that case, why don't you mark this as a draft (or close it) so that you don't feel the need to stay on top of it. We can come back to it once Glides has started publishing events and we formally document the schema.

paulswartz:
I've pushed up a version that's built with Docusaurus, and includes quite a bit of the work you've done here already.  I started you off with two of the events, and will leave it to you and Melody to finish the others.

https://mbta.github.io/schemas/events/glides/com.mbta.ctd.glides.editors_changed.v1
https://mbta.github.io/schemas/events/glides/com.mbta.ctd.glides.operator_signed_in.v1

PR #2, opened 2023-06-02
Schemas for Glides Events

skyqrose:
Re-opened version of #1 based on the new Docusaurus format.

Asana Task: [Glides Events RFC and Schema](https://app.asana.com/0/1200273269966439/1203911426224983/f)

This copies everything from [RFC 18](https://github.com/mbta/technology-docs/blob/main/rfcs/accepted/0018-glides-trainsheet-events.md#alternative-include-snapshot-of-all-data-not-just-changes) so those docs are now obsolete.

This PR has:
- Schemas:
  - Schemas for all 3 events that Glides publishes.
  - A shared `schemas/glides-events.json` which is the type of the union of all 3 events, and has some definitions that are shared among the 3 events.
  - References between the files.
  - Potentially some minor corrections compared to RFC 18, I don't quite remember.
- Examples and tests:
  - Examples, all in `examples/glides-events/`, broken apart by scenario, so the docs can reference each scenario.
  - Significant improvements to the validation script. Now the one script compiles and checks all the schemas and validates all the examples.
  - (The new examples are all in `glides-events/`, so they're validated against the big union `glides-events.json` instead of directly against their individual event schema.)
- Docs:
  - `index.mdx` docs files instead of categories, so we can write things on each index page. (This also removes `category/` from the url).
  - Upgrades to the `<Examples/>` component.
  - Docs for all 3 events, with:
    - description of the event
    - descriptions of the data types
    - examples with commentary
    - the embedded schema viewer
  - Docs at `docs/events/glides/index.mdx` with:
    - high level description about how Glides publishes events
    - links and short descriptions to all 3 events
    - descriptions of data types shared among the events

I recommend reviewing these files as a whole, rather than looking at their diffs:
- `schemas/glides-events.json`
- `schemas/com.mbta.ctd.glides.*.json`
- `src/scripts/validate_examples.ts`

I recommend reviewing everything in `docs/events/glides/*.mdx` by running it locally and seeing it in your browser rather than by reading the `.mdx` files.


boringcactus (commented):
The trips_updated schema viewer appears to fail to resolve the trip_updated and trip_added refs, and just shows as `any OR any`, at least when I build the docs locally.

paulswartz:
@skyqrose I'll take a look later, but I've updated the repo permissions: can you see if you're able to assign reviewers now?

skyqrose:
@/boringcactus Oh, that's gonna be hard to fix.

The problem is that while building static assets, and dereferencing all the `$ref`s into one flat schema to put in `build/schemas/`. The library creates one correctly dereferenced schema, but then we iterate through it to add back some detail from the original. For example, if we have the schema `"source": {"$ref": "cloudevents.json#/properties/source", "const": "glides.mbta.com"}`, we want to keep both the dereferenced version from cloudevents and the extra detail added in this schema, which the resolver doesn't do.

The current bug is that our custom code doesn't recurse into lists while doing that and that would be pretty easy to fix. But it also doesn't recurse into dereferenced objects, and there's not an easy way to do that with the webpack-merge approach, so we're basically going to have to use the library just for the resolver and then write our own dereferencer.

Other issues I found while poking around this problem:
- The dereferencer is written in js. With this change it'll be complex enough to be worth having typescript check stuff, but it's hard to have a typescript docusaurus plugin.
- The [json dereferencing library](https://github.com/stoplightio/json-ref-resolver) is deprecated. We should switch to [a newer one](https://github.com/APIDevTools/json-schema-ref-parser).
- The straightforward object merge leaves some weird artifacts. Like keeping `examples` from the dereferenced schema when the referencing schema specifies a specific `const`. We might want to do a JSON Schema specific merge.

That'll all be a lot of code to fix. I'm happy to do it, but it's unrelated enough to the current PR that I'd probably want to ignore the issue for now and then do it as a followup PR.

schemas/glides-events.json
```diff
@@ -0,0 +1,112 @@
...
+  "$ref": "cloudevents.json",
+  "type": "object",
+  "properties": {
+    "type": {
+      "type": "string",
+      "description": "name of the event",
+      "$comment": "enforced to be a valid event name by the oneOf, below"
+    },
+    "specversion": { "const": "1.0" },
+    "source": { "const": "glides.mbta.com" },
```
skyqrose: Note to self: This should probably be set differently per environment, instead of a const.

docs/events/glides/index.mdx
```diff
@@ -0,0 +1,165 @@
...
+
+Fields:
+
+- `badgeNumber` (string)
+
+MAY refer to other non-operator Area 132 employees in some cases, for example if an inspector signs in as fit for duty in Glides (they aren't required to).
+
+It is represented as an object to provide future extensibility if needed.
+
+## Details
```
paulswartz, line 114: question: this looks like it's pretty much copied from RFC 18, which is already linked above. Do you think there's additional value to re-including the content here?

docs/events/index.mdx
```diff
@@ -0,0 +1,13 @@
+---
```
paulswartz, line 1: praise: thanks for this! It looks much nicer.

docs/events/glides/com.mbta.ctd.glides.editors_changed.v1.mdx
```diff
@@ -1,35 +1,65 @@
...
+Fields in the event's `data`:
+
+- `metadata` ([Metadata](.#metadata)): how the changes were entered in Glides, including the author.
+- `changes` (array of [EditorChange](#editor-change)): a list of start and stop editing events.
+
+### EditorChange
+
+Fields:
+
+- `type` (string `"start"`|`"stop"`): Whether the editor started or stopped editing.
```
paulswartz, line 35: question: should this be the "description" in the schema? In general, this section seems like it's duplicating the information that's in the rendered schema UI.

schemas/glides-events.json
```diff
@@ -0,0 +1,118 @@
...
+  "$ref": "cloudevents.json",
+  "type": "object",
+  "properties": {
+    "type": {
+      "type": "string",
+      "description": "name of the event",
+      "$comment": "enforced to be a valid event name by the oneOf, below"
+    },
+    "specversion": { "const": "1.0" },
+    "source": {
```
paulswartz, line 15: suggestion: this is missing either a `type` or a `$ref` to something with a `type`.

docs/events/glides/com.mbta.ctd.glides.operator_signed_in.v1.mdx
```diff
@@ -1,29 +1,63 @@
...
+## Data
+
+Fields in the event's `data`:
+
+- `metadata` ([Metadata](.#metadata)): how the operator was signed in in Glides, including the inspector who signed them in.
+- `operator` ([Operator](.#operator)): the operator who signed in
+- `signedInAt` (RFC3999 timestamp): the time at which they signed in (this is separate from the event timestamp)
+- `signature` ([Signature](#signature)): how the operator signed in.
+
+### Metadata
```
paulswartz, line 24: question: does anything link to this? the `Metadata` above links directly back to the Glides root description.

docs/events/glides/com.mbta.ctd.glides.editors_changed.v1.mdx
```diff
@@ -1,35 +1,65 @@
...
+
+Note that the author may or may not match the editors in changes, since inspectors can change whether other people are editing.
+For example, when an inspector takes over editing at a location, they may sign out the previous editor.
+Then they would issue a stop change for the previous editor and a start change for themself.
+
+## Data
+
+Fields in the event's `data`:
+
+- `metadata` ([Metadata](.#metadata)): how the changes were entered in Glides, including the author.
```
paulswartz, line 28: nitpick: you don't need the leading `.`, and removing it avoids a page reload (I think)
```suggestion
- `metadata` ([Metadata](#metadata)): how the changes were entered in Glides, including the author.
```

docs/events/glides/com.mbta.ctd.glides.trips_updated.v1.mdx
```diff
@@ -0,0 +1,288 @@
...
+
+If fields are missing from ScheduledCar, it is not known who is scheduled to operate that car. An empty object is valid if we know the car is scheduled to operate but don't know anything else about it.
+
+### Time
+
+Time in the `HH:MM:SS` format. The time is measured from "noon minus 12h" of the service day (effectively midnight except for days on which daylight savings time changes occur). For times occurring after midnight, enter the time as a value greater than 24:00:00 in `HH:MM:SS` local time for the day on which the trip schedule begins. Effectively, a [GTFS Time](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#field-types) but without support for `H:MM:SS` times.
+
+> _Example: `04:30:00` for 4:30AM or `25:35:00` for 1:35AM on the next day._
+
+Alternative: times as RFC3999 string.
```
paulswartz: question: is this useful context for the schema documentation?

docs/events/glides/com.mbta.ctd.glides.trips_updated.v1.mdx
```diff
@@ -0,0 +1,288 @@
...
+
+Some values might be the special string `"unset"`. This means to treat the field as if it had never been modified. This is semantically slightly different than setting the value to the scheduled value. For example, setting `startTime` to `"unset"` implies we don't know when the trip will leave, and the scheduled time might be a good guess, but setting `startTime` to equal the scheduled start time implies that an inspector has confirmed the trip will leave at that time.
+
+_Notes:_
+
+- setting a new location or time does not modify the [TripKey](#tripkey) for a scheduled trip.
+- If `cars` is updated from a 1-element list to a 2-element list, then Glides SHOULD include data for the 2nd car. This is relevant if a two-car train has a car removed and then re-added. If a field was set before the car was dropped, then when the car is restored, clients MUST assume that the field is `"none"` as opposed to retaining its previous value, but Glides SHOULD include the data to avoid the ambiguity.
```
paulswartz, line 175: suggestion: in this documentation, `SHOULD` is probably not useful for consumers, as they can't rely on it. Does Glides provide this data, or not?

docs/events/glides/com.mbta.ctd.glides.trips_updated.v1.mdx
```diff
@@ -0,0 +1,288 @@
...
+            Due to staffing issues, Inspector Alice drops the 10:05am trip,
+            along with its return trip. (First event with two updated trips).
+          </p>
+          <p>
+            Using the "Manage Headways" feature, Alice changes the expected
+            departure times of the remaining trips to 9:56am, 10:02am, and
+            10:08am. (Second event with three updated trips).
+          </p>
+          <aside>
+            (Scheduled data is unrealistically abbreviated for this example.
```
paulswartz, line 218: suggestion: the first example also has abbreviated schedules, but does not have this note.

paulswartz (commented):
I reviewed some of the files, but in general, it seems to me like much of the textual description should be "title" or "description" values in the schema. This would having needing to specify the types and formats of the fields twice: once in the schema and again in the text.

docs/events/glides/index.mdx
skyqrose: Yes.
- The descriptions are closer to the schema viewer, schemas, and everything else.
- The examples here are verified and are more useful to reference when reading the descriptions, the examples in RFC 18 had errors.
- This repository will be kept up to date, the RFC won't.
- Everything's in one place. This is now a single source of truth. RFC18 can be safely forgotten.

docs/events/glides/com.mbta.ctd.glides.editors_changed.v1.mdx
skyqrose: I was wondering about that.

Reasons to put all the documenation in the schema:
- It can all be machine-readable and verified. As is this could have unnoticed errors. Later updates to the schema could especially cause errors.
- Stuff is all in one place. You don't have to jump between prose and schema to correlate the two.
- If someone fetches a schema file, they'd get the documentation too, instead of the documentation being only available on the docs website.

Reasons this is better here:
- Easier to navigate this organized document than clicking through the interactive schema viewer.
- Easier to read paragraphs than string fields.
- Easier to read human-readable descriptions than machine-readable. e.g. `"start"|"stop"` vs `"anyOf": [{"const": "start"}, {"const": "stop"}]`
- Can have richer text. e.g. links to other documentation.
- Can describe things in prose rather than associating each description with a specific field or object in the schema.
- Handling and maintaining docs is easier than handling and maintaining schemas. E.g. there are errors in the dereferencing code that means some documentation would have been clobbered if it were in the schema. Writing text is simpler and just works.

I ended up deciding on keeping the schema focused on validation, and putting the meaning and documentation here where it can be freely organized and formatted and read. But I could be convinced otherwise.

schemas/glides-events.json
skyqrose: Good catch. Fixed.

docs/events/glides/com.mbta.ctd.glides.operator_signed_in.v1.mdx
skyqrose: No, nothing links here, because if it did, it should just link directly to the shared glides-events page instead.

This is here so that if someone comes here looking for information about an operator_signed_in's metadata, they can find it, instead of it being hidden on another page. They could probably find the metadata field and follow that link, but in general, the way I organized these docs was that each event page gets all the object types that show up in it, and I didn't want an object type to be missing from this event's page just because it shows up on another event so was put on the shared page instead.

docs/events/glides/com.mbta.ctd.glides.editors_changed.v1.mdx
skyqrose: It is needed.
`#metadata` links to this page's `/events/glides/com.mbta.ctd.glides.editors_changed.v1.mdx#metadata`
`.#metadata` links to the parent page's `/events/glides/#metadata`

docs/events/glides/com.mbta.ctd.glides.trips_updated.v1.mdx
skyqrose: A little? I can imagine someone reading this and thinking "why are they using this weird format instead of just using normal timestamps?". This is there to reduce the confusion of that hypothetical reader.

(Though in this specific case, we might end up switching to timestamps because they're better for representing unscheduled overnight nonrevenue trains. TBD.)

docs/events/glides/com.mbta.ctd.glides.trips_updated.v1.mdx
skyqrose: Good catch. I've added more a more realistic schedule to that one.

(I added the data instead of the note because it's the first and most basic example so it's more important that it's realistic, and because it was only one car so it's not distracting from the rest of the data.)

docs/events/glides/com.mbta.ctd.glides.trips_updated.v1.mdx
skyqrose: Good point. I've gone through a reworded a couple instances of "Glides SHOULD" to be more certain where I know what we'll do.

But in some cases, like this one, I don't want to promise too much and I don't want clients to rely on it. "SHOULD" is still the right way to express the contract that we'll try to follow this, but we might not, and we want to be able to change the behavior if we have a good reason without clients breaking.

For reference, [RFC keywords definition](https://www.rfc-editor.org/rfc/rfc2119) for SHOULD:
> This word, or the adjective "RECOMMENDED", mean that there may exist valid reasons in particular circumstances to ignore a particular item, but the full implications must be understood and carefully weighed before choosing a different course.

Examples where it would be wrong to remove the SHOULD and imply we're certain, so I kept it:

- In TripAdded: "The event SHOULD only include values which were explicitly included by the author: inferred values SHOULD NOT be included." We currently have [a bug](https://app.asana.com/0/616151179860796/1204649263600516/f) where we include some inferred data. The data is still usable, though, and "the bug means we're misleading about how reliable this data is" is a less disasterous outcome than "the bug means we've broken this contract".

- In TripAdded "`dropped` SHOULD NOT be set". It would be weird if was set, but I can't promise that it would never happen. Like what if during a disruption where trains are running from an unscheduled endpoint, someone is adding trips as they go. They're adding trips at a regular headway, but then end up short an operator and can't run the next trip. They add the trip to the trainsheet to mark where they would have sent it, and immediately drop and write in a reason. Semantically, that's a trip that's both added and dropped at the same time and it's fine.

docs/events/glides/com.mbta.ctd.glides.trips_updated.v1.mdx
paulswartz: If you're going to keep it, can you make it clear that this was an alternative considered for the schema, and not an alternative that's a valid value for this field?

schemas/com.mbta.ctd.glides.trips_updated.v1.json
```diff
@@ -1,91 +1,33 @@
...
-      "required": []
-    },
-    "trips_updated_v1": {
-      "type": "object",
-      "properties": {
+        "metadata": { "$ref": "glides-events.json#/$defs/metadata" },
         "tripUpdates": {
           "type": "array",
-          "contains": {
+          "items": {
```
paulswartz, line 18: I think this is the correct change, but something else is resulting in the sub-objects not appearing in the rendered schema:
Screenshot 2023-06-14 at 3 09 29 PM

docs/events/glides/index.mdx
paulswartz: Can you add a note that the documentation here supersedes the descriptions in RFC 18?

schemas/com.mbta.ctd.glides.operator_signed_in.v1.json
```diff
@@ -4,21 +4,32 @@
...
   },
-  "required": ["type", "specversion", "source", "id", "time", "data"]
+  "required": ["type", "specversion", "source", "id", "time", "data"],
+  "$defs": {
+    "signature": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "type": "string",
+          "examples": ["tap", "type"]
```
paulswartz, line 30: question: did you mean for these to be `const` values instead? The documentation doesn't seem to indicate that this is a free-text field.

skyqrose: It is supposed to be free text so that any changes to how we represent signatures don't cause a breaking change, and because supporting arbitrary strings here isn't a big burden on a client. I've clarified that in the documentation now.

docs/events/glides/com.mbta.ctd.glides.trips_updated.v1.mdx
skyqrose: I've removed it.

schemas/com.mbta.ctd.glides.trips_updated.v1.json
skyqrose: I've traced the problem down to the way the schemas are dereferenced (when merging the dereferenced and original schemas, it doesn't recurse into list field). But there are other changes I want to make to how that dereferencing happens. Since that's pretty independent of the Glides schemas and documentation, I've made a followup task: https://app.asana.com/0/616151179860796/1204796800599703/f , and I'd open a separate PR for it. (I haven't finished fixing it yet, but I plan to.)

skyqrose:
Is this okay to merge?

There is still the bug with sub objects not getting dereferenced, which lead to the "any" problem above. But the schema source is correct, the bug is in the build script. I have [an asana task](https://app.asana.com/0/616151179860796/1204796800599703/f) for that and I'll fix it sometime soon. I don't think it should be blocking for this PR.

docs/events/glides/com.mbta.ctd.glides.trips_updated.v1.mdx
```diff
@@ -0,0 +1,288 @@
...
+- `scheduledCars` (array of [ScheduledCar](#scheduledcar)): Array of length 1 or 2. The length reflects the scheduled length of the train. In a 2 car train, the front car is listed first.
+
+It does not include time and location fields, because for scheduled trips those fields are already in [TripKey](#tripKey)
+
+### ScheduledCar
+
+Fields:
+
+- `run` (string, optional): The run number scheduled to the trip.
+- `operator` ([Operator](#operator), optional): The operator who is scheduled to operate the run that day, and therefore scheduled to do the trip.
```
paulswartz: question: should this link to `.#operator` instead? From the above section, it looks like there's no other documentation for this field here.

```suggestion
- `operator` ([Operator](.#operator), optional): The operator who is scheduled to operate the run that day, and therefore scheduled to do the trip.
```

docs/events/glides/index.mdx
```diff
@@ -0,0 +1,165 @@
...
+
+### Kinesis
+
+Events will be written as records to a Kinesis stream. Each Glides environment (`dev`, `dev-green`, and `prod`) will have a separate stream, named `ctd-glides-<environment>`.
+
+The partition key will be a hash of the station at which the inspector is working and the inspector's identity, which ensures that multiple events from a single inspector are ordered correctly if the records are distributed across multiple shards.
+
+Multiple events can be included in a single Kinesis record, by wrapping them in a JSON array. This is an optimization for improving write speeds: further steps in the Kinesis pipeline may break up or rearrange these arrays.
+
+Kinesis only adds a limited (<1s) latency between when the event is generated and when it is available to consumers.
```
paulswartz: suggestion: It looks like this is messing up the Github syntax highlighting: 
Screenshot 2023-07-07 at 8 44 14 AM
```suggestion
Kinesis only adds a limited (&lt;1s) latency between when the event is generated and when it is available to consumers.
```

paulswartz (commented):
a couple small things, but otherwise I think this is good to go!

PR #4, opened 2023-07-28
feat: schema for Subwayside

paulswartz:
You can see the live version at https://paulswartz.github.io/schemas/events/com.mbta.ctd.subwayside.train-heartbeat


schemas/com.mbta.ctd.subwayside.train-heartbeat.json
```diff
@@ -0,0 +1,394 @@
...
+    "carNumber": {
+      "type": "string",
+      "description": "The 4 digit number of a car",
+      "pattern": "^\\d{4}$"
+    },
+    "timestamp": {
+      "description": "RFC3339 timestamp",
+      "type": "string",
+      "format": "date-time",
+      "pattern": "^\\d{4}-(0\\d|1[012])-([012]\\d|3[01])T([01]\\d|2[0-3]):([0-5]\\d):([0-5]\\d)([.]\\d{1,6})?(Z|[+-]\\d{1,2}:[0-5]\\d)$"
```
beatrixada, line 56: question: is this validation pattern necessary? apparently JSON Schema already expects RFC 3339 datetimes: https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times

beatrixada (approved):
@paulswartz I notice most of the fields aren't documented. I don't personally care about that because it sounds like we really only need three of them for crowding, but do we plan to document them more extensively somewhere?

schemas/com.mbta.ctd.subwayside.train-heartbeat.json
paulswartz: `format` isn't required to be implemented by validators. From slightly above your link:
> By default, `format` is just an annotation and does not effect validation.

beatrixada:
Oh one last question: is there a specific reason so many fields are named `*Dt` etc instead of `*Date`?

beatrixada (commented):
question: why do some fields use abbreviations like `nbr` or `Dt`?

paulswartz:
> Oh one last question: is there a specific reason so many fields are named *Dt etc instead of *Date?

`*Dt` is for date-time. I'd probably have used `*Date` for actual dates, or `*Time` for times.

> question: why do some fields use abbreviations like nbr or Dt?

The fields are `nbr` or `*nbr` in the database, but I can change that. Some options that come to mind:
- `number` 
- `carNumber`
- `carId`

paulswartz:
@bklebe I've updated all the `Nbr` to `Number`, but kept `Dt`.

PR #5, opened 2024-03-22
feat (glides): add revenue field to trips_updated

skyqrose:
Asana Task: [Publish nonrevenue trips to RTR](https://app.asana.com/0/616151179860796/1206310370383903/f)

Currently, Glides only publishes events for revenue trips, and all trips in the event stream are assumed to be revenue trips.

This adds a `revenue` field to the schema so we can publish nonrevenue trips. It adds the field to the `TripKey` so that it's possible to tell whether the scheduled trip is a pull or a revenue trip, and it adds it to `TripAdded` and `TripUpdated` so that Glides can add nonrevenue trips and change whether a trip is in service.

The default if the fields don't exist (as is today) is that the trip is assumed to be in service, so there are no immediate changes needed for Glides or RTR.

We will check with the transit data team before publishing any nonrevenue trips.

Is it okay with the transit data team if Glides begins publishing `revenue: true` for revenue trips without notice? This should be backwards compatible if RTR doesn't do anything with the field yet.

### DIFF
--- a/schemas/com.mbta.ctd.glides.trips_updated.v1.json
+++ b/schemas/com.mbta.ctd.glides.trips_updated.v1.json
@@ -125,7 +125,7 @@
           "properties": {
             "revenue": {
               "$comment": "if revenue is missing when adding a trip, assume it's in revenue service. if revenue is missing when updating a trip, that means no change, so there is no default",
-              "default": true
+              "default": "revenue"
             }
           }
         }
@@ -152,7 +152,10 @@
             "endLocation": { "$ref": "glides-events.json#/$defs/location" },
             "startTime": { "$ref": "#/$defs/time" },
             "endTime": { "$ref": "#/$defs/time" },
-            "revenue": { "type": "boolean", "default": true }
+            "revenue": {
+              "oneOf": [{ "const": "revenue" }, { "const": "nonrevenue" }],
+              "default": "revenue"
+            }
           },
           "required": ["startLocation", "endLocation", "startTime", "endTime"]
         }
@@ -191,7 +194,7 @@
           "items": { "$ref": "#/$defs/car" }
         },
         "revenue": {
-          "type": "boolean"
+          "oneOf": [{ "const": "revenue" }, { "const": "nonrevenue" }]
         },
         "dropped": {
           "oneOf": [{ "$ref": "#/$defs/dropped_reason" }, { "const": false }]
--- a/schemas/com.mbta.ctd.glides.trips_updated.v1.json
+++ b/schemas/com.mbta.ctd.glides.trips_updated.v1.json
@@ -120,6 +120,14 @@
               "required": ["previousTripKey"]
             }
           ]
+        },
+        {
+          "properties": {
+            "revenue": {
+              "$comment": "if revenue is missing when adding a trip, assume it's in revenue service. if revenue is missing when updating a trip, that means no change, so there is no default",
+              "default": true
+            }
+          }
         }
       ]
     },
@@ -143,7 +151,8 @@
             "startLocation": { "$ref": "glides-events.json#/$defs/location" },
             "endLocation": { "$ref": "glides-events.json#/$defs/location" },
             "startTime": { "$ref": "#/$defs/time" },
-            "endTime": { "$ref": "#/$defs/time" }
+            "endTime": { "$ref": "#/$defs/time" },
+            "revenue": { "type": "boolean", "default": true }
           },
           "required": ["startLocation", "endLocation", "startTime", "endTime"]
         }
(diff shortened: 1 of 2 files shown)


schemas/com.mbta.ctd.glides.trips_updated.v1.json
```diff
@@ -181,6 +190,9 @@
           "maxItems": 2,
           "items": { "$ref": "#/$defs/car" }
         },
+        "revenue": {
+          "type": "boolean"
```
beatrixada: question: do you think we might want more trip types in the future, such that it would make more sense to make this an enum like `status` with values `revenue` and `nonrevenue`? It seemed plausible to me. e.g. what we did in the API: https://github.com/mbta/api/blob/f7b2efe4057f2261ab547c206289d32050061789/apps/api_web/lib/api_web/controllers/trip_controller.ex#L251-L260

skyqrose: oh, hm, I hadn't considered that.

I think "is this vehicle in revenue service?" could really be a boolean answer rather than an enum answer with two known options. There are many other statuses, but they could be separate fields.

Brainstorming future statuses we could add:
- "disabled"? This is something we'd be likely to add, but both revenue and nonrevenue vehicles could become disabled, so that should be a separate field rather than 4 statuses "rev disabled"/"rev nondisabled"/"nonrev disabled"/"nonrev nondisabled".
- type of nonrevenue vehicle, eg. wire car / instruction car. This could be useful, but there are many things you could do just knowing that a trip is nonrevenue without knowing why, so we should have a revenue field you can reference without knowing the full list of nonrevenue vehicle types, and then a separate vehicle type field.
- "charter"? This really would be a 3rd state that's neither revenue nor nonrevenue, but MBTA doesn't do that.
- "unknown"? It would be a 3rd state, but since this data all comes from human entry, right now we always know. But I can imagine future features that would create unknown data (e.g. automatically detecting a departure, and then trying but failing to match it up to a trip on the trainsheet, but we want to publish some other data about the trip).

okay, I'm convinced, I'll switch it to an enum.

beatrixada: Hmmm, the point about whether it's revenue or non-revenue is a good one. I suppose in many cases consumers will have to do this kind of categorization anyways even if there's several different types, so maybe it does make sense to have a separate field.

lemald: Isn't "disabled" kind of a fuzzy concept though, since a lot of different things can go wrong with a vehicle? It seems sort of like that's a separate concern from other status values, and if a vehicle isn't taking passengers due to being disabled you'd just mark is as non-revenue.

That said, this is mostly just bikeshedding since it was brought up, but I agree with the overall idea of having an enum.

skyqrose: Done!

nlwstein:
@skyqrose I have put up a PR that makes this change safe in RTR: https://github.com/mbta/rtr/pull/1281

Overall, I think your documentation here is very good. I am happy to approve once we get the pre-requisite code in place on RTR! 

My only question, which I believe may have already been debated, is why the enum over a boolean? If we already know of the other value(s) that it could be set to in the future, would it be worth documenting in this schema, so downstream applications can at least have those values as part of the data contract, if not having any specific behaviors tied to them?

skyqrose:
There aren't any specific other options, but it seems possible we might find others in the future. If we do, switching the field from a boolean to an enum would be a pain.

This is just for the published events. Internally in Glides (and potentially in RTR) the field is represented as a boolean, because changing the internal representation from a boolean to an enum in the future will only affect one codebase at a time and so would be relatively easy to refactor.

schemas/com.mbta.ctd.glides.trips_updated.v1.json
```diff
@@ -181,6 +193,9 @@
           "maxItems": 2,
           "items": { "$ref": "#/$defs/car" }
         },
+        "revenue": {
+          "oneOf": [{ "const": "revenue" }, { "const": "nonrevenue" }]
```
mathcolo, line 197: Are these references, or should they be `REVENUE` and `NONREVENUE` or neither?

mathcolo (approved):

schemas/com.mbta.ctd.glides.trips_updated.v1.json
skyqrose: We talked synchronously about this. The API uses the string constants "REVENUE" and "NON_REVENUE" ([example](https://api-v3.mbta.com/vehicles?filter%5Brevenue%5D=NON_REVENUE)), which I didn't realize when I arbitrarily chose the lowercase constants. I guess it'd be nice to use a consistent constant, but:
- It's unlikely anyone would be trying to join these data sources together.
- Consumers would probably parse the constant into an internal format anyway.
- Changing it isn't worth making RTR update their PR to use new constants.

So I'll leave the lowercase constants as is.

PR #6, opened 2024-04-18
feat (glides): small changes to TODS location ids

skyqrose:
Asana Task: [Publish nonrevenue trips to RTR](https://app.asana.com/0/616151179860796/1206310370383903/f)

Glides may start publishing data about nonrevenue trips to kinesis soon (not before Transit Data is ready). Those trips will contain new TODS stop ids.

A couple small changes to prepare:

1. ODS was renamed to TODS (_Transit_ Operational Data Standard). We might as well fix the name now while it's easy, lest we get locked into an out of date name forever.

2. The new location ids are likely to be a mix of location_type 0 and 1. (Inner Belt will be a parent of two children `ibeltm` and `ibeltu`. Other yards will be location_type 0 unless we have a reason to make them parents.) The schema used to guarantee that Glides wouldn't publish child stops, but now we say we might publish TODS child stops (but still not GTFS child stops).

<details>
<summary>These changes are technically backwards incompatible, but Glides hasn't published nonrevenue location ids yet and AFAIK nobody's looking at them, so I don't expect this to break anything.</summary>

Some pedantic exceptions:

Glides has tried not to publish any nonrevenue locations, but we technically could have in two situations:

- We publish `operator_signed_in` and `editors_changed` events with a location of inner belt.
- In the unlikely event an inspector changes a revenue trip into a nonrevenue trip, and then moves it to a yard, we could accidentally publish data about it.

In either case, we would have published `{"gtfsId": "an internal id that's not in GTFS"}`. We've never published `{"odsId": "anything"}` before.
</details>

This and #5 can go in either order.

### DIFF
--- a/docs/events/glides/index.mdx
+++ b/docs/events/glides/index.mdx
@@ -69,10 +69,8 @@ Fields:
 
 One of:
 
-- `{"gtfsId": "<GTFS stop ID>"}`, where the GTFS `location_type` is `1` (station).
-- `{"todsId": "<TODS stop ID"}`: id corresponding to `ops_location_id` in the upcoming TODS data. The data is still undergoing development and the specific ids are not yet finalized.
-
-Glides does not collect or publish any location data with GTFS `location_type` `0` (platform).
+- `{"gtfsId": "<GTFS stop ID>"}`, where the GTFS `location_type` is `1` (station). Glides doesn't differentiate between children of GTFS stops, so won't publish any GTFS child stop IDs.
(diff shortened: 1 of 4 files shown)


nlwstein:
This looks good to me, I have put up a PR to address this on RTR: https://github.com/mbta/rtr/pull/1284

We should probably deploy these changes at the same time, so there's minimal loss of data.

nlwstein (approved):

skyqrose:
No need to deploy at the same time, as we've never published this data before.
All of the `schemas` and `RTR` PRs can go in any order, and then the [Glides PR](https://github.com/mbta/glides/pull/2573/files) goes last.

mathcolo (approved):

PR #7, opened 2024-04-25
feat(glides): operator_signed_in's signature gets a version property

mathcolo:
Glides began tracking sign-in text version earlier this year, and we're planning on adding that to kinesis events.

### DIFF
--- a/docs/events/glides/com.mbta.ctd.glides.operator_signed_in.v1.mdx
+++ b/docs/events/glides/com.mbta.ctd.glides.operator_signed_in.v1.mdx
@@ -35,8 +35,8 @@ How the operator signed that they were fit for duty.
 
 Fields:
 
-- `type` (string): How the the operator made their signature.
-- `version` (integer): The version of the sign-in text the operator agreed to.
+- `type` (string, optional): How the the operator made their signature.
+- `version` (integer, optional): The version of the sign-in text the operator agreed to.
 
 Current values for `type` are
 
--- a/docs/events/glides/com.mbta.ctd.glides.operator_signed_in.v1.mdx
+++ b/docs/events/glides/com.mbta.ctd.glides.operator_signed_in.v1.mdx
@@ -36,6 +36,7 @@ How the operator signed that they were fit for duty.
 Fields:
 
 - `type` (string): How the the operator made their signature.
+- `version` (integer): The version of the sign-in text the operator agreed to.
 
 Current values for `type` are
 
--- a/examples/glides-events/operator_signed_in.v1.json
+++ b/examples/glides-events/operator_signed_in.v1.json
@@ -1,6 +1,6 @@
 {
   "type": "com.mbta.ctd.glides.operator_signed_in.v1",
-  "specversion": "1.1",
+  "specversion": "1.0",
   "source": "glides.mbta.com",
   "id": "19fdb184-7dd6-4664-8472-04bd6177ec44",
   "time": "2023-01-20T09:30:00-05:00",
--- a/examples/glides-events/operator_signed_in.v1.json
+++ b/examples/glides-events/operator_signed_in.v1.json
@@ -1,6 +1,6 @@
 {
   "type": "com.mbta.ctd.glides.operator_signed_in.v1",
-  "specversion": "1.0",
+  "specversion": "1.1",
   "source": "glides.mbta.com",
   "id": "19fdb184-7dd6-4664-8472-04bd6177ec44",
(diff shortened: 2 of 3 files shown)


examples/glides-events/operator_signed_in.v1.json
```diff
@@ -1,6 +1,6 @@
 {
   "type": "com.mbta.ctd.glides.operator_signed_in.v1",
-  "specversion": "1.0",
+  "specversion": "1.1",
```
paulswartz: suggestion!: `specversion` is the version of the CloudEvents specificiation, and it MUST be `1.0`: https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#specversion

docs/events/glides/com.mbta.ctd.glides.operator_signed_in.v1.mdx
```diff
@@ -36,6 +36,7 @@ How the operator signed that they were fit for duty.
 Fields:
 
 - `type` (string): How the the operator made their signature.
+- `version` (integer): The version of the sign-in text the operator agreed to.
```
paulswartz: question: Is the expectation that version is now a required field? If so, this would require a new event, `com.mbta.ctd.glides.operator_signed_in.v2`.

paulswartz: question: would it be more useful to store this as a string, or are we confident that we'll never have versions like "1.1" or "1a"?

schemas/com.mbta.ctd.glides.operator_signed_in.v1.json
```diff
@@ -28,6 +28,10 @@
         "type": {
           "type": "string",
           "examples": ["tap", "type"]
+        },
```
paulswartz, line 31: question: I missed this before, so this might be more of a question for @skyqrose, but this schema doesn't enforce that `type` is required, which doesn't appear to match the textual description above.

examples/glides-events/operator_signed_in.v1.json
mathcolo: whoa- thanks!

docs/events/glides/com.mbta.ctd.glides.operator_signed_in.v1.mdx
skyqrose: I think this should be an optional field. It would mean the change is backwards compatible so we don't need to bump the schema version, and the event is still very meaningful without the text version.

skyqrose: The Glides database keeps it as an integer, and we don't do any kind of semver or anything on the text, it's just if the text ever changes the number goes up. I think an integer is fine.

examples/glides-events/operator_signed_in.v1.json
skyqrose: For reference, the way to bump the event version is to make a new event like `com.mbta.ctd.glides.operator_signed_in.v2`, then we publish both events until all consumers have migrated, then we can stop publishing the `.v1` event ([docs](https://github.com/mbta/schemas/blob/47ce0baae62cbea14871fc4014830861e32e19e8/docs/events/glides/index.mdx#compatibility)). (but I don't think we need to do that here.)

docs/events/glides/com.mbta.ctd.glides.operator_signed_in.v1.mdx
paulswartz: In that case, can we update both of these to "(&lt;type&gt;, optional)" to match our other documentation?

mathcolo: oh, I assumed optional was the default. will update 👍

schemas/com.mbta.ctd.glides.operator_signed_in.v1.json
skyqrose: Oh, I think that's fine to be optional, it's still meaningful to say an operator signed in even if you don't know how they signed. But it does seem like an oversight that `event.data.signature` and `event.data.signedInAt` are required.
Preston, can you fix the schema to make `signature` and `signedInAt` required, and update the docs to make sure that things all the signature data is correctly labeled as required or optional?

mathcolo: Yep!

skyqrose (approved):

PR #9, opened 2024-11-14
feat(glides): new event vehicle_trip_assignment.v1

skyqrose:
Status as of Thursday 17:00: Made some edits, ready for a (maybe final?) review. If I've addressed a comment, please resolve the thread so it's easier to see what's left.

As proposed in the [Glides Trip Matching RFC](https://github.com/mbta/technology-docs/blob/main/rfcs/accepted/0025-light-rail-trip-model.md#glides-train---trip-assignment-events).
Asana Task: [Define schema for vehicle_trip_assignment events](https://app.asana.com/0/1200689710863995/1208467668804806/f)

[Rendered](https://www.skyqrose.com/mbta-schemas/events/glides/com.mbta.ctd.glides.vehicle_trip_assignment.v1)

This includes some fields that aren't strictly necessary. The spec could be shorter without them. However, I think they're useful, and should be included.
- `tripKey.serviceDate`: Trip IDs are repeated on consecutive days. You can probably figure out which date it's on to by the current time, but this is more explicit and better matches to [`trips_updated` `TripKey`](https://mbta.github.io/schemas/events/glides/com.mbta.ctd.glides.trips_updated.v1#tripkey).
- `tripKey.scheduled`: Distinguishes whether a trip is added or scheduled. Without it, consumers wouldn't be able to tell apart scheduled trips that happen to not be in their schedule, and added trips which would never match their schedule. That's not critical for RTR but should make consuming a little easier. It also makes matching to `trips_updated` `TripKey` easier, and discourages trying to parse the trip ID.
- ~~`tripKey.revenue`: This is definitely data about the trip, and not its ID, so isn't necessary. Consumers should be able to look up data about the trip in another way, and handle trips with unknown data. However, it's the one most important piece of data about the trip, could enable consumers to inform riders about a nonrevenue train even if they don't recognize the trip ID, and might help to avoid accidentally treating a nonrevenue train as if its revenue.~~
- ~~`vehicle.carNumbers`: The vehicle is uniquely referred to by the RTR `vehicleId` field, which RTR will use when it consumes the data, so this isn't necessary for identifying the vehicle. However, it will make it much easier to be able to use the data in other ways, and will make the feed way more human-readable for debugging.~~

~~The only nested list that will be difficult to stuff into a relational format will be `carNumbers`, but that's small enough that it shouldn't cause issues getting the data into LAMP.~~


docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+If `VEHICLE 1` was last assigned to `TRIP A`, and a new event assigns `VEHICLE 1` to `TRIP B`, Glides MUST also send another `VehicleTripAssignment` in the same event's batch to assign the `vehicle` for `TRIP A` to `null` or to a different vehicle. Consumers are not expected to deduce that `TRIP A` no longer has a vehicle assigned based on an update to `VEHICLE 1` and `TRIP B` (which doesn't mention `TRIP A`). Glides MUST ensure that after processing all updates in an event's batch, if the last update to `TRIP A` assigned `VEHICLE 1` to it, then `VEHICLE 1` MUST NOT have any other more recent updates.
+
+Conversely, if `VEHICLE 1` was last assigned to `TRIP A`, and a new event assigns `VEHICLE 2` to `TRIP A`, Glides MUST also assign `VEHICLE 1` to another trip or to a `null` trip in the same batch. By the end of processing all of the updates in each batch, if the last update to `VEHICLE 1` assigned it to `TRIP A`, then `TRIP A` MUST NOT have any other more recent updates.
```
skyqrose: I ended up with a couple awkward rules here to prevent consumers' state from getting tangled up. I'm happy with the current design, but if anybody finds the state-updating rules to be overly complex, here could be an alternative to discuss:

Instead of assigning vehicles or trips to null, we could do `{vehicle: v, trip: t, type: "assign"}` and `{vehicle: v, trip: t, type: "unassign"}`, and Glides has to unassign a pair before reassigning either of them.

Problems with the current spec:
- It's a bit complex.
- It says that you only end up in a good state after you finish processing the whole batch of assignments.
- If Glides messes up, or while a batch is partially applied, you could have a vehicle assigned to a trip, but the trip not assigned to the vehicle.

This alternative's pros:
- State is valid after every single assignment object, not just after the whole batch
- Removes the nullable fields and a couple edge cases.

This alternative's cons:
- Still requires the producer (Glides) to send updates in just the right order, and not send invalid updates.
- Is a bit backwards-facing (what was the old assignment, so we can undo it), while the current written spec is more forwards-facing (ignore the past, what's the train doing right now?).
- Could still end up in an invalid state if Glides messes up or if a consumer misses an event.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+The added/scheduled distinction is according to the schedule in Glides. If Glides has a different schedule than RTR or GTFS, then a trip that Glides thinks is scheduled may not match up with scheduled trip in GTFS, and RTR (or other consumers) may decide that a trip that was added in Glides lines up close enough to correspond to a scheduled trip in GTFS.
+
+Revenue is included because this is the most important information that Glides can provide to riders about a train, and its presence will make it less likely for a consumer of this feed to incorrectly interperet a nonrevenue train as being in service. Especially since nonrevenue trips are not in GTFS and their IDs are likely to not be recognized by a consumer.
+
+Consumers SHOULD use the use the entire object when comparing two `TripKey`s to see if they're equal, not just the `tripId` field. Otherwise, updates could be applied to the wrong trip, for example, updating the vehicle assignment for yesterday's trip instead of today's.
```
lemald: **nitpick:**
```suggestion
Consumers SHOULD use the entire object when comparing two `TripKey`s to see if they're equal, not just the `tripId` field. Otherwise, updates could be applied to the wrong trip, for example, updating the vehicle assignment for yesterday's trip instead of today's.
```

examples/glides-events/vehicle_trip_assignments.v1.json
```diff
@@ -0,0 +1,125 @@
+[
+  {
+    "type": "com.mbta.ctd.glides.vehicle_trip_assignments.v1",
+    "specversion": "1.0",
+    "source": "glides.mbta.com",
+    "id": "19fdb184-7dd6-4664-8472-04bd6177ec44",
+    "time": "2024-11-14T09:50:00-04:00",
```
paulswartz: nitpick: our current offset is `-05:00`

examples/glides-events/vehicle_trip_assignments.v1.json
```diff
@@ -0,0 +1,125 @@
...
+          "tripKey": null
+        }
+      ]
+    }
+  },
+  {
+    "type": "com.mbta.ctd.glides.vehicle_trip_assignments.v1",
+    "specversion": "1.0",
+    "source": "glides.mbta.com",
+    "id": "19fdb184-7dd6-4664-8472-04bd6177ec44",
```
paulswartz: question: is the expectation is that these IDs are supposed to be unique, so that this ID should be different from the event before it?

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+If there are no trains operating on any trips, for example, in the middle of the night after service ends, then for each `vehicleId` and `TripKey` in the feed, the last update to that vehicle or trip will be assigning it to a `null` trip or vehicle.
+
+Notes:
+
+Consumers MUST tolerate vehicles and trips being unassigned (assigned to `null`) even when the consumer didn't know they were previously assigned. This is needed because consumers might not have seen a previous update where they were assigned, as described in the docs on [Short-term storage](index.mdx#short-term-storage): "Clients MUST tolerate receiving updates for events that reference unseen past events". Updates that assign a vehicle or trip to a non-null trip or vehicle are unlikely to cause problems for consumers, though if consumers didn't see all past events, they might not know which trip or vehicle the vehicle or trip is being removed from.
+
+Glides MAY publish an event that assigns a vehicle to a trip even if the last update for that vehicle and trip assigned them to each other. Glides might do this to update the `carNumbers` associated with a `vehicleId`.
```
lemald: **suggestion:** I understand the rationale for including car numbers as a sort of reference for humans looking at the data, but I worry that this sort of implies that Glides would update the consist for a vehicle by publishing new events. I think it should be specified that Glides won't necessarily do this.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+Glides MUST NOT refer to the same vehicle or trip more than one time within all the assignments in one event.
+
+## Rules for Assignments
+
+This schema provides a method for saying which trains are assigned to which trips, but does not describe rules for when Glides will say a train is assigned to a trip. This section is a non-normative description of how Glides intends to decide which assigments to publish, but Glides may decide to change these rules and that would not be a breaking change to this schema, even if it has semantic consequences for consumers.
+
+This section may not be kept up to date as Glides' logic is updated.
```
lemald: **suggestion:** if it's not guaranteed that this section will be updated as the Glides logic changes, and if it's not as normative as other parts of the documentation, I wonder if it would make more sense to place this somewhere else like a Notion document (and then include a link to it from here).

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
```diff
@@ -0,0 +1,184 @@
...
+### TripKey
+
+Fields:
+
+- `serviceDate` (RFC3999 date): the service date for the trip.
+- `tripId` (string): ID, unique within the service date. Trip IDs for scheduled trips are intended to match GTFS trip IDs, but may not.
+- `revenue` (`"revenue"` | `"nonrevenue"`): Whether the trip is a revenue trip.
+- `scheduled` (`"scheduled"` | `"added"`): Whether this is a trip that Glides has in its schedule or that Glides added.
+
+This is a different structure than the [`TripKey` used in `trips_updated`](com.mbta.ctd.glides.trips_updated.v1.mdx#tripkey). Added and scheduled trips use the same format, and are differentiated by the `scheduled` field instead of by using different formats.
```
paulswartz: suggestion: since this is a different format from the other kind of TripKey, should it get a different name? Maybe `AssignedTripKey`?

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+For added trips, Glides makes up a new ID, which is kept consistent for any subsequent updates to the same added trip. Glides tends to prefix trip IDs for added trips with the string `"ADDED-"` but consumers MUST NOT assume any particular format, or assume that they can tell an ID for an added trip apart from a scheduled trip ID by the string alone. That's what the `scheduled` field is for.
+
+The added/scheduled distinction is according to the schedule in Glides. If Glides has a different schedule than RTR or GTFS, then a trip that Glides thinks is scheduled may not match up with scheduled trip in GTFS, and RTR (or other consumers) may decide that a trip that was added in Glides lines up close enough to correspond to a scheduled trip in GTFS.
```
paulswartz: question: given this, is it important for this event to make a scheduled/added distinction? Either the `tripId` is something RTR knows about or it's not, but is whether Glides thought it was scheduled needed for RTR?

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+Revenue is included because this is the most important information that Glides can provide to riders about a train, and its presence will make it less likely for a consumer of this feed to incorrectly interperet a nonrevenue train as being in service. Especially since nonrevenue trips are not in GTFS and their IDs are likely to not be recognized by a consumer.
```
paulswartz: nitpick:
```suggestion
Revenue is included because this is the most important information that Glides can provide to riders about a train, and its presence will make it less likely for a consumer of this feed to incorrectly interpret a nonrevenue train as being in service. Especially since nonrevenue trips are not in GTFS and their IDs are likely to not be recognized by a consumer.
```

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
paulswartz: question: is there a possibility that the other data changes (such as `revenue` for an added trip) but we'd want consumers to treat it as the same trip? should we be more explicit that the key is `{service ID, trip ID}`?

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
paulswartz: +1 to Eddie's suggestion.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
lemald: I think this is related to [my question about car numbers](https://github.com/mbta/schemas/pull/9#discussion_r1843922483) as well - are we saying that there may be events that differ only in fields like revenue or car numbers and should be treated as duplicates, or are we saying that the consumer should apply updates to those respective fields even if the trip key / vehicle ID combo stays the same?

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+## Updating Assignments
+
+Consumers SHOULD maintain state about which vehicle-trip assignments are active, and update that state when they receive an assignment that refers to a `vehicleId` or `tripKey`.
+
+Vehicles can only be assigned to one trip at a time. A vehicle that was previously assigned to `TRIP A` MAY be assigned directly to `TRIP B` without a separate assignment that assigns its trip to `null` first. Conversely, a trip can only have one assigned vehicle at a time, and doesn't need a previous vehicle to be removed before having a new vehicle assigned to it.
+
+If `VEHICLE 1` was last assigned to `TRIP A`, and a new event assigns `VEHICLE 1` to `TRIP B`, Glides MUST also send another `VehicleTripAssignment` in the same event's batch to assign the `vehicle` for `TRIP A` to `null` or to a different vehicle. Consumers are not expected to deduce that `TRIP A` no longer has a vehicle assigned based on an update to `VEHICLE 1` and `TRIP B` (which doesn't mention `TRIP A`). Glides MUST ensure that after processing all updates in an event's batch, if the last update to `TRIP A` assigned `VEHICLE 1` to it, then `VEHICLE 1` MUST NOT have any other more recent updates.
```
lemald: **question:**

> Consumers are not expected to deduce that `TRIP A` no longer has a vehicle assigned based on an update to `VEHICLE 1` and `TRIP B` (which doesn't mention `TRIP A`).

Why don't we want to place this expectation on clients? It seems like straightforward logic to implement, and this seems to add complications to the schema itself. In particular, I don't love that it means that it matters whether two assignment events are in the same batch event as opposed to coming immediately one after the other in separate batches.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
lemald: It feels like a lot of the complication here is because this schema as-written is trying to handle both vehicle to trip and trip to vehicle assignments, when the main underlying goal is vehicle to trip assignments and I'm not sure that we need to explicitly spell out the reverse direction. If we want clients to be able to deduce the reverse mapping, we can just spell out the semantics to indicate that upon receiving a Vehicle A -> Trip 1 mapping, it should clear out any other Vehicle X -> Trip 1 mappings in its state.

> ...requires the producer (Glides) to send updates in just the right order, and not send invalid updates.

Are there actually any barriers to Glides sending these events in the right order?

> Could still end up in an invalid state if Glides messes up or if a consumer misses an event.

What sort of failure modes would cause either of these things to happen?

lemald: Overall I'm more of a fan of the alternative, but I think it can be simplified even further:

> Glides has to unassign a pair before reassigning either of them.

Can't we just make the semantics of processing the `{vehicle: v, trip: t, type: "assign"}` message that the consumer has to remove any vehicle -> trip mappings involving either vehicle `v` or trip `t` before inserting the new mapping from `v` to `t`? You'd still have `"unassign"` events for when a vehicle is being unassigned from a trip without a replacement trip, or maybe you could skip `type` entirely and make `trip` nullable.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
lemald: I ended up actually making a pretty similar point in [your thread about the "alternative approach"](https://github.com/mbta/schemas/pull/9#discussion_r1844127293), so it might make sense to just put your combined response there.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+If Glides expects a train to turn around at a station in the middle of the line at a station where trains must loop to a different platform (Government Center Eastbound to Westbound, North Station, Kenmore, Park Street), when the train reaches the last stop, Glides will leave the train assigned to the previous trip. Glides will not unassign it from the previous trip and assign the train to the next trip until the train has actually turned around onto track that goes the other direction. This is so that if a train is unexpectedly extended (e.g. continuing from Government Center until North Station), the train won't be unassigned from a trip until it _actually_ finishes the trip.
```
lemald: **nitpick:** I think it would actually be safe to do the unassign-upon-arrival approach for trains looping at Park Street because once a train is on the fence track there it physically has to turn around.

lemald: (That's a weird edge case kind of nitpick, though, just something I wanted to point out without saying that means we should necessarily change the approach.)

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+### Other
+
+Glides may unassign or reassign vehicles to trips at other times, too. For example, if a train unexpectedly takes a different branch, or if a train turns around in an unexpected place, or an inspector enters new information into Glides. This section is not a comprehensive spec of how Glides decides which trips to assign vehicles to.
+
+Glides may not be able to make an assignment for every train. If it doesn't recognize the trip a train is doing, it will leave the train unassigned to any trip. It will not create a new added trip just to have a trip ID to reference. Trains will only be assigned to added trips if an official has written an added trip into the trainsheet in Glides.
+
+Glides may publish different assignments in the Glides app and in this event feed, if it decides the needs of operations staff (who use the Glides app) and riders (who use data powered by this feed) are different. We expect to follow slightly different rules around layovers for in-app trip assignments.
```
lemald: **question:** maybe out of scope, but do you think that there are any things we can say about the relationship between what shows up in the feed and what shows up in the app? I agree that in general there's no guarantee that they be 100% the same, but do you think we could say (for instance) that _if_ a train is assigned to a trip in the Glides UI, _then_ it will be assigned to that trip in the feed, but the converse isn't necessarily true? Or do you think that even that is too strong of an assertion to make?

examples/glides-events/vehicle_trip_assignments.v1.json
skyqrose: 87d1d80

examples/glides-events/vehicle_trip_assignments.v1.json
skyqrose: yep, I just got lazy when writing test data. They're unique now in 87d1d80

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: Good point. Except Glides currently ignores the child stop id and only looks at the parent station, so we don't know if a train is on the fence track, and it'd be a bit of work to change that. I guess the same might apply at Brattle  loop, too.

paulswartz: If Glides is only looking at the parent station, how does it know that the train has turned around?

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: The two sources might assign a vehicle to two different trips, not just one has an assignment and the other doesn't, so that specific assertion wouldn't work.
We're only planning to do it at layovers, so could maybe the spec could say that?

But the whole point of this sentence is that it's _not_ making any specific assertions. They should both be correct from their own perspective, and they'll be mostly the same, so it's not like a specific assertion here is needed for correctness. And it shouldn't affect consumers of the feed at all, except for humans looking at the feed and the app at the same time. So I think I don't want to make it more specific.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: It's not _required_, but it's useful:
- This field determines whether to match against `tripId` (scheduled) or `glidesId` (added) when matching to the `trips_updated` feed. Without it, RTR has to search both added and scheduled trips.
- Separating it means that the namespaces for added IDs and scheduled IDs are separate. We don't have to worry about what if Glides autogenerates an integer ID for an added trip that happens to collide with a scheduled ID.

I think of scheduled IDs and added IDs as two separate namespaces/types that happen to both be strings. It's more explicit and predictable to label them as a discriminated union, rather than to mix them into one field and then try to sort them out again after.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
lemald: I guess I agree that it doesn't make sense for these two different values to line up all the time, but I'm just thinking that from an architectural and overall legibility perspective it would make sense to put some very high-level guardrails around how much this feed can differ from the Glides app view. I don't want us to end up in a situation where Glides is the only application doing light rail trip assignments but it's actually doing multiple radically different sets of trip assignments for different purposes.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: Even though this section is non-normative, it does have a purpose for being part of the schema document:
- It's important for understanding the purpose of the schema. These assignments aren't arbitrary, they're generated by Glides based on the best data it has.
- Having a rough understanding of how layovers work will make working with the data much easier, since that's where most assignments and unassignments will happen, and where most developer attention will be, even if the rules can't be spelled out precisely.
- There are a number of details in here that are important for consumers to know because they affect how to develop features that use the stream (avoiding circular dependencies, assignments might happen at weird times, assignments might not match in-app data), even if they're not spelled out in strict rules. It's not just a copy of background context.
- This is written from the perspective of the schema, and wouldn't make sense if pulled alone into a standalone notion document.

This suggestion does make me see how this section is a bit out of place, though. Here are some ideas that might address that. They can be mix and matched.

1. As, suggested pull this section into a notion document.
2. Pull it into a notion document with significant edits, e.g. by rewriting it from Glides' perspective instead of the schema's. Try to be more specific about what Glides will do, so it makes more sense as a standalone document.
3. Simplify it. This document doesn't need a list of which stations have which type of layover. Just say "sometimes a layover will be reassigned immediately and sometimes it won't". Just enough detail to support the goals of the section listed above.
4. Move some rules to other normative parts of the spec. "Glides MAY make assignments at any location, not just terminals" with uppercase "MAY".
5. Move the general context about how this schema is an output of Glides trip matching and feeds into RTR into the very top of the document.

and of course every list of options should include

0. No build, leave it as it is.

I'd lean towards a combination of 3, 4, and 5. Or maybe 2, 4, and 5, but I don't know if we're ready to write real documentation for Glides' trip matching logic yet.

And, for reference, here's the [notion document](https://www.notion.so/mbta-downtown-crossing/Expected-Behavior-Sidebar-126f5d8d11ea8053a1f3e052ac1926fd#126f5d8d11ea8068b3d4d20b0dbc73f7) that I based some of this logic on, but it's not meant to be documentation, it's more like a set of notes, and written while thinking through a different problem.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: How could the spec avoid making that implication? Maybe in the `Vehicle` section, say something like "for context only, it isn't meant to be an authoritative source for car numbers, use GTFS-RT from RTR instead if you want to know what cars are associated with a vehicleId?"

Or maybe it'd be easier to just lean in to it and guarantee that we send updates for car numbers if they change? The car numbers come from GTFS-RT, so should be just as accurate as other sources, as long as Glides promises to keep them up to date.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: Oooh, I had never considered that we wouldn't need an explicit "trip -> vehicle" mapping, too. And that implicitly clearing state like that would be fine to ask clients to do. Like, it shifts logic from the producer to the consumer, which I normally don't like, but it does remove some edge cases in the spec, so I think it's worth it.

> Are there actually any barriers to Glides sending these events in the right order?

No, I'm just pointing out that the alternative doesn't eliminate illegal states.

> What sort of failure modes would cause either of these things to happen?

If a Glides bug results in an invalid message (e.g. reassigning a second vehicle to the same trip), or if a client misses a message and gets the state out of sync so that a valid message _looks_ like it's reassigning a second vehicle to the same trip, then the client might not have a good way to handle the event that leaves it in a legal state. (This wouldn't happen at start up from missing old history. It would only happen if a client receives some events, then misses an event, then receives another event.) This is easy to avoid by just not writing any bugs, but I'm pointing it out as a flaw in both the current version and alternative I proposed. A well-designed spec would always have a defined way to handle any incoming message.

Actually, I think your last idea, where both vehicles and trips can both be implicitly unassigned, could end up with no messages that are illegal to handle from any state, and updates can be processed one at a time instead of in batches. This might be the ticket. I'll sleep on it over the weekend and try writing it up on Monday.

skyqrose: A followup question: You suggested only vehicle to trip assignments (vehicle is not null, trip might be null). What if it was the other way around, and the feed was trip to vehicle assignments (trip is not null, vehicle is nullable. For each trip, which vehicle is on it?)

This doesn't match as well with RTR's use case, where it has a vehicle and wants to know its trip. But it would better fit with the `trips_updated` feed, where `tripKey` is the id, and car numbers are the data.

Or maybe the schema could be symmetrical where neither one of them needs to be the primary key, either of them could be null, and the client can use either one as a key depending on their use? I think the benefits from this discussion come from the implicit unassignment in either direction, not from focusing on vehicle to trip assignments.

Again, I'll try it out Monday and see how it goes.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: Ooh, good question, and I guess this problem is one of the tradeoffs in including the revenue field even though it's not required. I'm not sure.

For scheduled trips, revenue is part of the schedule and shouldn't change (and if it does, that's specified in trips_updated), so I'm fine treating it as part of the ID.

For added trips, it's more likely for the revenue status to change. In that case it's more like a payload than the ID. But this is a bad format for publishing updates to trip data. If revenue was part of the ID, then if it was ever changed, then you'd have two different added trips with the same `{serviceDate, tripId, scheduled: "added"}`, which could cause problems if you weren't consistent about always including revenue in the ID.

But if added trips didn't have the revenue field, then if a consumer hadn't seen that trip come through trips_updated, then they'd have no idea if the vehicle was in revenue service or not.

I think including revenue in the ID so the whole TripKey is the ID is the most straightforward answer, so I'm leaning that way, but I'm not certain.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: By the direction we get from RTR. (Which, awkwardly, is a property of the trip assignment, hence the note about circular dependencies in the "Inputs" section.)

paulswartz: @lemald or @Whoops  maybe you can chime in about how we think this will work on the RTR side? From an outsider perspective, it seems like the direction is part of the trip assignment, so Glides would need to be responsible for it (and couldn't rely on it as an input to the trip assignment).

lemald: If I recall correctly, the direction assignment at RTR actually happens at a lower level of the logic than the trip assignment, the issue is just that trip details are the place where the direction ID actually flows through in the GTFS-rt, so it's more of an issue of how the data is formatted than a fundamental circular dependency in the logic.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: I did 3 in [0801c8d](https://github.com/mbta/schemas/pull/9/commits/0801c8df48ba418c024c778f861a7e3675042eae), still deciding on how/whether to do any others.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
lemald: Looking at the updates I like that it's simplified! I guess we could go either way on the directionality, but I think there's benefit in choosing one direction and sticking with it, just because otherwise if there's an active assignment of Vehicle A to Trip 1, then there are actually two different ways to unassign it with the same semantics (assigning Vehicle A to trip null or assigning vehicle null to Trip 1).

skyqrose: Did the implicit unassignments in [00399b4](https://github.com/mbta/schemas/pull/9/commits/00399b406442269fc470c6f9e1aa0fe3e96f3c85). It's definitely better.

I haven't changed the symmetry between vehicles and trips. The main benefit to making it asymmetrical would be that consumers don't need to handle either vehicle being null OR trip being null. But it wouldn't change the state that consumers need to save, or simplify how updates are made. I'm leaning towards leaving it as symmetrical because it makes less promises about non-null fields, so is more forwards compatible.

lemald: > I'm leaning towards leaving it as symmetrical because it makes less promises about non-null fields, so is more forwards compatible.

How much forward compatibility does this actually buy us? If we allow trip ID to be null, allowing null vehicle IDs with non-null trip IDs just gives us a different way of expressing something that could already be expressed with a non-null vehicle ID and a null trip ID, and if we wanted it to mean something different that would be a breaking change in the semantics anyway.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
lemald: > Or maybe it'd be easier to just lean in to it and guarantee that we send updates for car numbers if they change?

For what it's worth, I would strongly prefer we not go this direction, and instead just make it explicit that the only time one of these messages is guaranteed to be generated is when the trip assignment state changes.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: It'd matter if we ever want to give other information about a trip, and want to be able to give that information even when there's no vehicle assigned. If null vehicles are allowed, then we can start publishing `{tripKey: {key}, vehicle: null, foo: "whatever"}` and it would work in a totally backwards-compatible way. If we ban null vehicles, then we have to change the spec and update all consumers before we start publishing this data.

Asymmetric nonnull vehicles would also implies that vehicles-to-trips is the preferred way to view the data. It's the way that RTR would use it, but I don't think it's inherent. After all `trips_updated` uses trips as keys and does trip-to-vehicle matching. Another consumer could use trips as keys for their matching state if that makes sense for their use case.

So it's just hypothetical benefits, but the cost (adding a null case) is small. Keep in mind that RTR will already have to write a `unassignVehiclesFromTrip(state, tripKey)` function anyway, because it could be triggered by `{vehicle: {nonnull}, tripKey: {was assigned to another vehicle}}`. This is just about whether you also need a `{vehicle: null, trip: tripKey} ->` case to call that existing function.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: Resolved in [00399b4](https://github.com/mbta/schemas/pull/9/commits/00399b406442269fc470c6f9e1aa0fe3e96f3c85)

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: +1. Conceptually, the direction of a train can be defined by which track it's on and which way it last moved. That's important input information into the trip matching logic. If we want to remove the circular dependency in trip assignments, then we would need a separate non-trip direction field in GTFS-RT for RTR to communicate a train's direction to Glides.

We can completely prevent the logical circular dependency with these rules:
- Glides can use the direction from GTFS-RT's trip, but nothing else.
- RTR never uses Glides as input when determining a vehicle's direction. If it gets a trip assignment from Glides where the trip's direction doesn't match the direction it expects, then it ignores Glides' assignment and makes its own trip with the correct direction.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
lemald: > Another consumer could use trips as keys for their matching state if that makes sense for their use case.

Sure, but then they could do that by maintaining a vehicle to trip mapping based on these events and then reversing it.

> It'd matter if we ever want to give other information about a trip, and want to be able to give that information even when there's no vehicle assigned.

Why would we put that other information in a `vehicle_trip_assignment` event as opposed to just creating a new event?

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
lemald: > If we want to remove the circular dependency in trip assignments, then we would need a separate non-trip direction field in GTFS-RT for RTR to communicate a train's direction to Glides.

Yes, I really like this idea, or maybe even separating it out further from GTFS-rt if we do some refactoring of RTR that allows it to expose more low-level state of things like vehicles to other consumers.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
```diff
@@ -0,0 +1,176 @@
...
+
+Does not provide predictions for which vehicles will be assigned to future trips.
+
+See the the [Updating Assignments](#updating-assignments) section for how vehicles and trips are assigned and unassigned.
+
+## Data
+
+Fields in the event's `data`:
+
+- `assignments` (array of [VehicleTripAssignment](#vehicletripassignment)): The list of vehicles and trips with new assignments.
```
lemald: **question:** how great do we feel the benefit is of including an array of `assignments` as opposed to each event containing a single assignment update? It would save us some bytes when Glides updates multiple vehicles - trip assignments at once, which I guess could happen when new realtime data comes in, but assignment updates from trainsheet edits would probably tend to come one at a time. I think especially if we've reworked the semantics such that the state will be consistent after processing each individual element of the assignments array, we should at least consider just making the structure flat.

Whoops: For my 2c, I'd prefer a single event per. That frees us up on the receiving end to process them one by one vs a batch of n, and I think makes more sense thinking of this from an event-driven perspective.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
```diff
@@ -0,0 +1,176 @@
...
+
+This is a different structure than the [`TripKey` used in `trips_updated`](com.mbta.ctd.glides.trips_updated.v1.mdx#tripkey). Added and scheduled trips use the same format, and are differentiated by the `scheduled` field instead of by using different formats.
+
+- If `scheduled` is `"scheduled"`, then this TripKey's `tripId` will equal the `trips_updated` TripKey's `tripId`.
+- If `scheduled` is `"added"`, then this TripKey's `tripId` will equal the `trips_updated` TripKey's `glidesId`.
+- (In both cases, the `serviceDate` must also match in order to know the two TripKeys refer to the same trip.)
+
+The difference with the TripKey in `trips_updated` is because `trips_updated` has to give enough information about a trip that a consumer can understand it even without recognizing the trip ID, while `vehicle_trip_assignments` assumes that you can look up information in `trips_updated` or in `GTFS`, so doesn't give as much information as possible about the trip. (Though consumers should keep in mind that order of events between two event streams is not guaranteed, and a trip that's referenced in this feed may not have appeared in `trips_updated`.)
```
Whoops: Am I following this correctly that we need to look at both `scheduled` and `trip_id` to know what field of `trips_updated` to match this to?

In the event of an added trip, what will the `trip_id` of `trips_updated` be? It would be nice if there was a single shared identifier.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
Whoops: From my perspective, I haven't been imagining RTR second-guessing Glides. We might have to deal with the case where we can't match a Glides trip_id to GTFS, but if Glides tells us that a trip is unscheduled/added, I don't think we should rerun trip matching.

Even as far as the scheduled case goes, I'm inclined to trust unless there turns out to be a reason I'm not thinking of we need to look up a trip in GTFS. If RTR were to second-guess glides and do it's own matching, then RTR might as well just do it.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
```diff
@@ -0,0 +1,176 @@
...
+
+Revenue is included because this is the most important information that Glides can provide to riders about a train, and its presence will make it less likely for a consumer of this feed to incorrectly interpret a nonrevenue train as being in service. Especially since nonrevenue trips are not in GTFS and their IDs are likely to not be recognized by a consumer.
+
+Consumers SHOULD use the entire object when comparing two `TripKey`s to see if they're equal, not just the `tripId` field. Otherwise, updates could be applied to the wrong trip, for example, updating the vehicle assignment for yesterday's trip instead of today's.
+
+The `revenue` and `scheduled` fields use string enums instead of booleans so that other states could be added in the future. Clients MUST tolerate values for these strings that they don't recognize. A safe way to handle this would be to treat the TripKey like an unrecognized trip, or to treat it like it's `null` i.e. the trip doesn't exist and a vehicle assigned to it is treated like it's unassigned from any trips.
```
Whoops: Do you have thoughts on what other states we might want in these fields at a later date? It's easy enough to be tolerant, but "Is this on the schedule" or "is this revenue generating" feel like binary questions to me, and RTR treats them as such.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: 👍 We'll make them as a batch when we get new vehicle positions from GTFS-RT, but also one by one at other times. And the issue about requiring batches in order to be consistent is fixed. If it'll be easier for you, then I'll go ahead and make this change.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: You need to look at `scheduled` to see which field to compare to, and `trip_id` to see which value to look for there.

In an added trip, `trips_updated` doesn't have a `trip_id` field, the `TripKey` is `{serviceDate, glides_id}`. ([docs](https://mbta.github.io/schemas/events/glides/com.mbta.ctd.glides.trips_updated.v1#added-trip))

It would be nice to have a single field to compare to. That's a flaw in the `trips_updated.v1` spec, because it was written before we started using GTFS ids. Now I'm dreaming of a `trips_updated.v2` that combines the two `TripKey` types into one and has a single `trip_id` field, but that would be a separate change, and this `vehicle_trip_assignments.v1` spec would be forward-compatible with that change.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: For "scheduled", I could imagine a "added, but inferred by Glides as opposed to entered by an inspector" state, or a "in Glides' HASTUS schedule but we already checked and know it won't be in GTFS" state.

For "revenue", I can't think of any third states, but this matches the enum used in [`trips_updated`](https://mbta.github.io/schemas/events/glides/com.mbta.ctd.glides.trips_updated.v1#scheduled-trip).

In both cases, it's mostly an attempt to support future states we _haven't_ thought of yet. They're definitely binary now, and Glides also treats them as booleans, but if the situation arises, it's easy to update applications to change the type to an enum, and much harder to change the events. Doing it this way to start with saves us a couple dance moves in a hypothetical future migration.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: Done in [0079c2b](https://github.com/mbta/schemas/pull/9/commits/0079c2bdbf488b46dd4e5f1db0f94d74e730a971)

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: By "match" in my comment, I meant looking up the data about the one trip in GTFS or the trip_updates feed. I didn't mean that RTR would do its own trip assignment logic.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
```diff
@@ -0,0 +1,167 @@
...
+
+For added trips, Glides makes up a new ID, which is kept consistent for any subsequent updates to the same added trip. Glides tends to prefix trip IDs for added trips with the string `"ADDED-"` but consumers MUST NOT assume any particular format, or assume that they can tell an ID for an added trip apart from a scheduled trip ID by the string alone. That's what the `scheduled` field is for.
+
+The added/scheduled distinction is according to the schedule in Glides. If Glides has a different schedule than RTR or GTFS, then a trip that Glides thinks is scheduled may not match up with scheduled trip in GTFS, and RTR (or other consumers) may decide that a trip that was added in Glides lines up close enough to correspond to a scheduled trip in GTFS.
+
+Revenue is included because this is the most important information that Glides can provide to riders about a train, and its presence will make it less likely for a consumer of this feed to incorrectly interpret a nonrevenue train as being in service. Especially since nonrevenue trips are not in GTFS and their IDs are likely to not be recognized by a consumer.
```
skyqrose: If `revenue` is removed, how would RTR determine whether a trip is in revenue service or not? What if the trip isn't in GTFS and hasn't shown up in `trips_updated`, what would RTR do?

Is there any other information about a trip besides revenue that's not derivable from realtime sources and would make a big difference to what data RTR publishes?

lemald: > What if the trip isn't in GTFS and hasn't shown up in `trips_updated`, what would RTR do?

Why would it not have shown up in `trips_updated` if it had already been marked as non-revenue? I can understand maybe a veeeery slight delay between one message and the other if some trainsheet edits happen roughly simultaneously (say someone changed both the revenue status and the car numbers in one edit), but I don't think that would be enough lag to create much of a problem.

lemald: I'll echo [my comment on the thread about which fields should be nullable](https://github.com/mbta/schemas/pull/9#discussion_r1847104346): these events are for providing information about trip - vehicle assignments, not for providing other information about trips. I think we're really over-complicating this by trying to make the `vehicle_trip_assignment` events do things that other events defined in the schema already do.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
Whoops: Where is Glides getting vehicle information in the current version? This has moved around a bit, and I'm not sure where we've settled on at the moment.

If the plan is to get it from an RTR feed rather than reading the trike feed, I don't think there's too much value in car numbers. It does save a human a round trip, but RTR will already have logged them, and the vehicle ID will be consistent between both, so it's really only for if the human happens to be reading the raw events.

On the other hand, if Glides is reading Trike directly, this could be a handy sanity check.

Overall, I think I lean on the side of leaving consist out. I don't like the idea of having data in the event that may go stale, if it's there I kind of feel like we should keep it up-to-date, but issuing a new TripAssignment for consist doesn't make sense semantically.

If we did keep it, I think I'd argue for broadening the scope, where this is something like a VehicleUpdate rather than a TripAssignment each event be the full state of a vehicle (it's assigned to + consist + any other information we wanted), where any change to vehicle information could trigger it, but I doubt rethinking this PR in that direction is appealing to anyone.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
Whoops: Not a strongly held belief, but I find myself leaning towards `vehicle` being non-nullable. The main reason for that is allowing either to be nullable means there's multiple way to accomplish the same thing. If I need to unassign vehicle 1 from trip 2, I can either send `%{vehicle: 1, trip: null}` or `%{vehicle: null, trip: 2}`. Picking vehicle as the key forces us to choose a lane and reduces the needed handling code on the consumer side.

That said, if, for some reason, the implementation is easier on the glides side, if we also support the trip-oriented version as well (or instead!) I'm not inherently opposed. But from the limited amount I've seen, Glides seems somewhat vehicle-oriented (Eddie or Sky, please correct me if I'm off base on that) so having looked at exactly zero code my guess is that the vehicle-oriented version is probably what Glides would end up sending in most situations anyway.

skyqrose: Good point about having multiple ways to do things. I even ran into that while writing examples.
My only remaining hangup is future compatibility if we ever want to add data to a trip without a vehicle, but I guess that the `trips_updated` feed handles that.

👍 Okay, so I'll make vehicle nonnullable.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: It will be some amount of effort to look up car numbers when debugging, but that data does exist in a few sources. The cost from the complexity will be worse than the extra debugging effort, so I'll remove it.

I'll also un-nest vehicle_id into the top level, removing the vehicle object.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: Vehicle is now nonnullable in 454856f

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: After talking with Eddie and Walton, I've removed the revenue field in 22245e1.

RTR should get the revenue status (and other trip info) from the trips_updated feed. RTR should log if they fail to find data there.

If it becomes a problem, we could reinstate the revenue field, though it should probably be a top-level field within `data` rather than part of the `TripKey`. (In that case, it'd be more like a field of the vehicle "is the vehicle accepting passengers" rather than a field on the trip.)

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: Another way that Glides might do different matching in-app vs this feed: We could consider showing inferred matches (where inspectors haven't entered data) in this feed, but consider those trains unmatched in-app, in order to prompt inspectors to manually fill out the trip and to avoid showing operator assignments we're less certain of.

(We aren't currently planning to do this, but it's something we've talked about and I wouldn't want the spec to prohibit it.)

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: revenue was removed in 22245e1. Now it's just the `scheduled` field that's the enum.

Also "inferred" wouldn't be a good example of a new state because that should be part of the data, not part of the trip ID. But "automatically created placeholder" and "scheduled but not in GTFS" might still be flavors of ID.

I'd still like to keep the enum, but less strongly than when we also had `revenue`.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: As discussed on the call, revenue was removed (22245e1). If RTR has trouble finding info in trips_updated (because of the delay, or because the trip was never edited), it should log that, and we could consider adding revenue back (as a data field, not as part of the `TripKey`).

Glides should try to ensure that any trip updates get sent before this event, so the trip can be referenced when RTR consumes this event.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: I've been pondering this. I agree, it is confusing that they're the same. But `TripKey` is a good name, and the two versions mean the same thing even if they accomplish it in different ways. And if we ever make a breaking change to `trips_updated`, we may want `trips_updated.v2` to use this version of `TripKey`, in which case `AssignmentTripKey` would be misleading.

I haven't thought of any better synonyms for `TripKey` that describes this type. `TripId` means just the string, not this object. `TripDescriptor` in GTFS-RT also includes some data fields that this doesn't. `Trip` implies having data, not just a reference.

I'm currently thinking that having two different definitions of `TripKey` really is the best way to represent this, but I'm still trying to think of other options.

docs/events/glides/com.mbta.ctd.glides.vehicle_trip_assignments.v1.mdx
skyqrose: If we need to ensure that the trip exists in the `trips_updated` feed before sending any `vehicle_trip_assignment`s, I had an idea for one way we could do that: https://github.com/mbta/schemas/pull/10

lemald: Also, for the record, this is only really an issue for _scheduled_ non-revenue trips, because those are the ones that RTR wouldn't have via GTFS or via TripUpdates (barring any edits made to them). However, the only scheduled non-rev trips we really have at the moment that go on what's normally revenue trackage and therefore need to be suppressed are Inner Belt pullouts / pullbacks, and there is already a mechanism for suppressing those. Also, in the long run, RTR can learn about scheduled non-revenue trips via TODs.

lemald (approved):
Okay, there are probably always more things we could discuss, but at a high level my outstanding questions about the basic schema have all been resolved. 👍

fsaid90 (approved):
No comments beyond what was already brought up/resolved - thank you!

paulswartz:
@skyqrose a question that came up in my 1-1 with Logan this morning: how does this work for future assignments? For example:
- vehicle 123 is currently doing trip A
- the next trip on the block is trip B
- however, Glides knows (because of inspector input) that the vehicle is going to be moved to a new block and run trip C instead

How is that going to be represented?

skyqrose:
@paulswartz That's not part of this events feed. RTR, when making predictions, can predict which vehicle will do a future trip based on the block or based on the `trips_updated` feed.

Whoops (approved):

paulswartz:
Does the `trips_updated` events have a vehicle identifier? I see that it has `cars` but not `vehicleId`. Without that, I'm not sure how RTR would know that the following trip for the vehicle has been changed. Maybe we can include an example about how it works if I'm missing something?

lemald:
> Does the `trips_updated` events have a vehicle identifier? I see that it has `cars` but not `vehicleId`. Without that, I'm not sure how RTR would know that the following trip for the vehicle has been changed. Maybe we can include an example about how it works if I'm missing something?

Vehicle ID would be preferable, but when an inspector is filling out car numbers, the corresponding vehicle may not exist in the live tracking yet. In any event RTR is aware of both the car numbers and the vehicle ID and can use the car numbers to find the vehicle in its state, if it is in fact tracking.

PR #10, opened 2024-11-21
docs(glides): trips_updated could publish schedules via no-op events

skyqrose:
This is an idea for how Glides could publish its schedule to RTR: As a `trips_updated` that mentions a trip but does not edit it.

There's been some chatter about whether Glides should publish its schedule at the beginning of the day, like OCS does, and this is one way that it could do that.

There was also a minor issue in https://github.com/mbta/schemas/pull/9 : If a vehicle is assigned to a scheduled trip that has never been edited, but the trip isn't in GTFS, then it's possible that RTR wouldn't be able to find information about the trip. One workaround to that could be to publish one of these events before publishing the `vehicle_trip_assignment`.

This PR adds a new example that follows the existing spec with no changes. It's possible this could just work with no special handling from RTR. Though Glides would need some work to decide when to send these events, and if we make important decisions about when to send events like this, that should be documented in a real normative way, not just as a hypothetical example.

This PR is mostly a way to share the idea and spark discussion. It could be merged as is, but it's not important to merge it and it's also not important that we start publishing events like this any time soon.


lemald:
FWIW, on the other PR I wrote out [some reasons why I don't believe this will end up being necessary](https://github.com/mbta/schemas/pull/9#discussion_r1853898996).

skyqrose:
Won't merge this or seek more feedback on it until we have a reason to do it, but if we ever do, we can reopen this.

PR #12, opened 2025-03-12
feat(events): add com.mbta.rtr.lightrail-vehicle-information

rudiejd:
Asana ticket: [[extra]🔮 Define schema for light rail vehicle locations feed for Glides](https://app.asana.com/0/584764604969369/1209274901226039/f)

Problem:
Glides needs information about light rail vehicles to start doing trip assignment on its own. [Further outlined here.](https://www.notion.so/mbta-downtown-crossing/Glides-Trips-RTR-Inter-dependency-Problem-Statement-185f5d8d11ea80d5a7b5e8c24cb85ce7)

Solution:
Add `com.mbta.rtr.lightrail-vehicle-information` message type with all of the necessary information for Glides LRV trip assignment.


schemas/com.mbta.rtr.light-rail-vehicle-information.json
```diff
@@ -0,0 +1,100 @@
...
+          "type": "string",
+          "examples": ["Green-D", "Green-C", "Mattapan"]
+        },
+        "consist": {
+          "description": "Array containing car numbers of the train in order",
+          "type": "string",
+          "examples": [["3644", "3862"]]
+        },
+        "orientation": {
+          "description": "Whether the 'rear' or 'front' of each train is facing forward for each train in the consist. B means backward, a means forward",
```
heybenji: a capital "A" might help clarify this statement. And, is it expecting pairs? And are they trains or cars?

schemas/com.mbta.rtr.light-rail-vehicle-information.json
```diff
@@ -0,0 +1,100 @@
...
+          "type": "string",
+          "examples": ["AB", "BA", "BB"]
```
heybenji: Maybe different examples than the "orientation" property above?

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
```diff
@@ -0,0 +1,22 @@
+---
+---
+
+import LightRailVehicleInformation from "@site/examples/com.mbta.rtr.light-vehicle-information/LightRailVehicleInformation.json";
```
skyqrose: I tried to run this locally. For some reason I'm getting this error:

> Module not found: Error: Can't resolve '@site/examples/com.mbta.rtr.light-vehicle-information/LightRailVehicleInformation.json' in '/{...local path...}/docs/events'

But I can't figure out why. If it works on your machine, then don't worry about it.

schemas/com.mbta.rtr.light-rail-vehicle-information.json
```diff
@@ -0,0 +1,100 @@
...
+        "consist",
+        "orientation",
+        "nextStopId",
+        "stopStatus",
+        "lat",
+        "lon"
+      ],
+      "properties": {
+        "directionId": {
+          "description": "Direction in which the vehicle is headed. Derived from the direction in trips.txt: https://gtfs.org/documentation/schedule/reference/#tripstxt",
```
skyqrose: The docs generation doesn't copy these descriptions into the body text. They're in the schema component, but hidden behind dropdowns, cluttered with other UI, and could be broken if [this bug](#3) bites.

UI of schema for OCS messages, showing a field description buried beneath dropdowns.

I recommend putting descriptions in the `.mdx` docs, rather than here.

schemas/com.mbta.rtr.light-rail-vehicle-information.json
```diff
@@ -0,0 +1,100 @@
...
+          "type": "integer"
```
skyqrose: This could have `"enum": [0, 1]`, cuz not every integer is a valid direction.

[docs](https://json-schema.org/understanding-json-schema/reference/enum) [example](https://github.com/mbta/schemas/blob/4c580eac6ae205311e5675ec153f2565a9881d38/schemas/com.mbta.ctd.subwayside.train-heartbeat.json#L273)

schemas/com.mbta.rtr.light-rail-vehicle-information.json
```diff
@@ -0,0 +1,100 @@
...
+        },
+        "vehicleId": {
+          "description": "Unique identifier of the vehicle from OCS",
+          "type": "string",
+          "examples": ["G-10023"]
+        },
+        "routeId": {
+          "description": "Route that RTR thinks the vehicle is serving. Value is from GTFS trips.txt: https://gtfs.org/documentation/schedule/reference/#tripstxt",
+          "type": "string",
+          "examples": ["Green-D", "Green-C", "Mattapan"]
```
skyqrose: Instead of examples, this could be comprehensive and used for validation with `enum: ["Green-B", "Green-C", "Green-D", "Green-E", "Mattapan"]`

schemas/com.mbta.rtr.light-rail-vehicle-information.json
```diff
@@ -0,0 +1,100 @@
...
+        },
+        "consist": {
+          "description": "Array containing car numbers of the train in order",
+          "type": "string",
```
skyqrose: This needs to be `type: "array", items: { type: "string" }`

schemas/com.mbta.rtr.light-rail-vehicle-information.json
skyqrose: Are the A and B ends as "forward" and "backward"? I'd always thought of the cars as symmetrical, and the labels arbitrary.

Does "AB" mean the first car has A forward and the second car has B forward, or does it mean that one car has A in front of B? In GTFS-RT multi_carriage_details, each car gets its own direction (either "AB" or "BA"), rather than the whole train having one field. One field for the train is fine, but if it's different from GTFS-RT that should be intentional and documented.

schemas/com.mbta.rtr.light-rail-vehicle-information.json
skyqrose: I think you meant to put this comment on `"nextStopID: { "examples": ["AB", "BA", "BB"] }` below?

schemas/com.mbta.rtr.light-rail-vehicle-information.json
```diff
@@ -0,0 +1,100 @@
...
+        },
+        "nextStopId": {
+          "description": "Identifier for the next stop that the vehicle is at / approaching. From GTFS stops.txt: https://gtfs.org/documentation/schedule/reference/#stopstxt",
+          "type": "string",
+          "examples": ["AB", "BA", "BB"]
+        },
+        "stopStatus": {
+          "description": "What the train is doing in relation to the next stop",
+          "type": "string",
+          "examples": ["IN_TRANSIT_TO", "INCOMING_AT", "STOPPED_AT"]
```
skyqrose: Could use an enum again.

schemas/com.mbta.rtr.light-rail-vehicle-information.json
```diff
@@ -0,0 +1,100 @@
...
+          "examples": [["3644", "3862"]]
+        },
+        "orientation": {
+          "description": "Whether the 'rear' or 'front' of each train is facing forward for each train in the consist. B means backward, a means forward",
+          "type": "string",
+          "examples": ["AB", "BA", "BB"]
+        },
+        "nextStopId": {
+          "description": "Identifier for the next stop that the vehicle is at / approaching. From GTFS stops.txt: https://gtfs.org/documentation/schedule/reference/#stopstxt",
+          "type": "string",
```
skyqrose: For all the fields in this file: Are any of them nullable? I like non-nullable fields, but if any are supposed to be nullable, in JSON Schema you'd have to specifically say they are with `"type": ["other type e.g. string", "null"]`.

For all the string fields: A common bug in producers is that a string is empty when it shouldn't be. If this schema is ever used for validation, you could hedge against that by putting `"minLength": 1` on any string fields that shouldn't be empty.

schemas/com.mbta.rtr.light-rail-vehicle-information.json
```diff
@@ -0,0 +1,100 @@
...
+          "examples": ["IN_TRANSIT_TO", "INCOMING_AT", "STOPPED_AT"]
+        },
+        "lat": {
+          "description": "The latitude of the train, provided by the AVL",
+          "type": "string",
+          "examples": ["42.21469"]
+        },
+        "lon": {
+          "description": "The longitude of the train, provided by the AVL",
+          "type": "string",
```
skyqrose: woah, lats and lons are strings? I guess that makes sense if you want to do exact comparisons. But Glides would just immediately parse these into numbers.

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
```diff
@@ -0,0 +1,22 @@
...
+import EventSchemaPath from "@site/src/components/EventSchemaPath.tsx";
+import Examples from "@site/src/components/Examples.tsx";
+
+Message from RTR indicating information about the location and status of light rail vehicles provided by RTR. RTR sources this information from the OCS and/or the vehicle AVLs.
```
skyqrose: Would RTR ever delete a vehicle, or just let it go stale? If it is deleted, how?

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
rudiejd: interesting, it does work on my machine (and seems to pass CI). I'm wondering if you're on a different yarn version or something?

schemas/com.mbta.rtr.light-rail-vehicle-information.json
rudiejd: I didn't think the [type system for CloudEvents](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#type-system) supports double/float, hence the string.

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
rudiejd: yes, if OCS sends us a lat/lon of 0.0/0.0 for a train then we delete it. You can [see this happening in prod here](https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22rtr-prod%22%20%22delete%20msg%20for%20train%22&display.page.search.mode=fast&dispatch.sample_ratio=1&workload_pool=&earliest=-30m&latest=now&sid=1741879088.60544). Would you want some mechanism by which RTR informs Glides that the vehicle is being deleted, or is that being handled elsewhere?

schemas/com.mbta.rtr.light-rail-vehicle-information.json
skyqrose: Oh! TIL. JSON Schema supports floats, but I see CloudEvents is trying to be more strict than JSON Schema so that it can also be serialized to non-json formats. 👍

schemas/com.mbta.rtr.light-rail-vehicle-information.json
rudiejd: yes, our [vehicle struct](https://github.com/mbta/rtr/blob/86824e5f326ed0625d00245095d510b5b372ddad/lib/rtr/vehicles/vehicle.ex#L53-L72) has almost every field as nullable for better or worse 😓 

would you like us to only send you vehicles in which all of these fields are populated? we could always filter on the RTR side

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
skyqrose: Oh, there is a typo, and it's not [getting tested in CI](https://github.com/mbta/schemas/pull/12/checks#step:4:40).
> skipping examples/com.mbta.rtr.lightrail-vehicle-information/, no matching JSON Schema in schemas/com.mbta.rtr.lightrail-vehicle-information.json

This line says `light-vehicle`. The examples directory is `lightrail-vehicle`. And the event name is `light-rail-vehicle`.

rudiejd: ahhh, good catch! I was just running `yarn ci` locally not the actual application. `yarn start` does not work for me.

schemas/com.mbta.rtr.light-rail-vehicle-information.json
rudiejd: ah, I might have understood this incorrectly. I thought that "AB" meant that one car was facing in the "A" direction and one car was facing in the "B" direction. I admittedly don't really understand this field well, and didn't realize that it was a GTFS-RT field. I was basing my idea of what the field should be on RTR's [ocs_messages.md](https://github.com/mbta/rtr/blob/master/docs/ocs_messages.md) and the [code that we have in RTR](https://github.com/mbta/rtr/blob/204a4e4e65237fa23fc970eebebe761986fb4e59/lib/ocs/parser/tmov_message.ex#L168-L187) for parsing `orientation`.  My assumption was since we get one OCS message per consist, the orientation must have to include information about all cars in a consist. do you know where `orientation` is specified in the GTFS-RT? I only see `carriage_sequence` in our docs. But I do see that the value is being populated for every car in a consist instead of at the consist level
image
image

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
skyqrose: Currently, Glides handles it by completely replacing our list of vehicles every time we poll. But with this event stream sending updates one vehicle at a time, we wouldn't be able to do that, so there'd need to be an explicit way to update a vehicle to be deleted.

schemas/com.mbta.rtr.light-rail-vehicle-information.json
skyqrose: Ah sorry, it's not in the GTFS-RT standard, but RTR adds it to the GTFS-RT enhanced json file. So the prior art I was suggesting you follow is just whatever the Transit Data Team made up last time, and it'd be fine to make something different up this time.

But whatever you choose, in a 2-car train it does have to be able to describe the orientation of both the front car and the rear car. (and it has to work on 1-car trains, too)

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
rudiejd: would glides be able to discard updates for vehicles that aren't in its list of vehicles? Otherwise, we could  send another notification with something like `isDeleted` when we get a TmovDel

schemas/com.mbta.rtr.light-rail-vehicle-information.json
rudiejd: I opened a PR to add this information to our GTFS docs for the enhanced feed: https://github.com/mbta/gtfs-documentation/pull/58. Thank you for pointing this out.

It seems like we could do something like 

```
"consist": [
    {
        "id": "3646",
        "orientation": "AB"
    },
    {
        "id":  "3892",
        "orientation": "AB"
     }
]
```


We could even rename the field to `multi_carriage_details` if we prefer consistency with GTFS-RT

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
skyqrose: I think the goal is for this to replace the existing polling, so it has to be able to communicate deletions within the event stream.

@mathcolo and @cho-jos should weigh in on how they want to know a vehicle was deleted.

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+Related to [RFC-25](https://github.com/mbta/technology-docs/blob/main/rfcs/accepted/0025-light-rail-trip-model.md)
+
+## Data
+
+Fields in the event's `data`:
+
+- `directionId` (integer): Direction in which the vehicle is headed. Derived from the direction in trips.txt: https://gtfs.org/documentation/schedule/reference/#tripstxt
+- `lastAvi` (string): The last AVI code that was emitted from the vehicle. If you are internal to the MBTA, see the documentation on AVI codes here https://github.com/mbta/rtr/blob/f734e361cb7361ec0cde461add8e1e1bfc938b74/docs/avi_codes.pdf
+- `vehicleId` (string): Unique identifier of the vehicle from OCS
```
mathcolo: Looking at realtime dashboard's "GTFS Trip Assignments" page, does this correspond to OCS trip ID or vehicle ID? "vehicleId" sounds like the latter, but "from OCS" sounds like the former.

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
```diff
@@ -0,0 +1,39 @@
...
+- `routeId` (string): Route that RTR thinks the vehicle is serving. Value is from GTFS trips.txt: https://gtfs.org/documentation/schedule/reference/#tripstxt
+- `consist` (array[string]): Array containing car numbers of the train in order
+- `orientation` (enum, values: ["AB", "BA"]): Whether the 'rear' or 'front' of each train is facing forward for each train in the consist. B means backward, a means forward
```
mathcolo: "a" => "A"?

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
```diff
@@ -0,0 +1,39 @@
...
+- `nextStopId` (string): Identifier for the next stop that the vehicle is at / approaching. From GTFS stops.txt: https://gtfs.org/documentation/schedule/reference/#stopstxt
+- `stopStatus` (enum, values: ["STOPPED_AT", "INCOMING_AT", "IN_TRANSIT_TO"]): What the train is doing in relation to the next stop
+- `lat` (string): The latitude of the train, provided by the AVLThe latitude of the train, provided by the AVLThe latitude of the train, provided by the AVL
```
mathcolo: This description is duplicated/typoed a couple of times?

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+Fields in the event's `data`:
+
+- `directionId` (integer): Direction in which the vehicle is headed. Derived from the direction in trips.txt: https://gtfs.org/documentation/schedule/reference/#tripstxt
+- `lastAvi` (string): The last AVI code that was emitted from the vehicle. If you are internal to the MBTA, see the documentation on AVI codes here https://github.com/mbta/rtr/blob/f734e361cb7361ec0cde461add8e1e1bfc938b74/docs/avi_codes.pdf
+- `vehicleId` (string): Unique identifier of the vehicle from OCS
+- `routeId` (string): Route that RTR thinks the vehicle is serving. Value is from GTFS trips.txt: https://gtfs.org/documentation/schedule/reference/#tripstxt
+- `consist` (array[string]): Array containing car numbers of the train in order
+- `orientation` (enum, values: ["AB", "BA"]): Whether the 'rear' or 'front' of each train is facing forward for each train in the consist. B means backward, a means forward
+- `nextStopId` (string): Identifier for the next stop that the vehicle is at / approaching. From GTFS stops.txt: https://gtfs.org/documentation/schedule/reference/#stopstxt
```
mathcolo: At one point @lemald and I discussed whether or not Glides should have access to numeric station IDs as a mechanism for Glides to derive more information about track movement. I know that's _not_ what we're doing here, but would it be better for this to be a child ID in case we want it later for an unrelated purpose?

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
mathcolo: Could deletions be in another key at the same level as `data`, or does that violate CloudEvents? I don't love `isDeleted`, as wouldn't it contain superfluous/old data?

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
rudiejd: this is a child stop ID. the child stop IDs are also in `stops.txt`

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
rudiejd: This is the Vehicle ID from OCS. That page shows the assignments from OCS trip => GTFS trip (and will show the assignments from Glides trip => GTFS trip for `dev-green` right now)

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
rudiejd: > yes, if OCS sends us a lat/lon of 0.0/0.0 for a train then we delete it. You can [see this happening in prod here](https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22rtr-prod%22%20%22delete%20msg%20for%20train%22&display.page.search.mode=fast&dispatch.sample_ratio=1&workload_pool=&earliest=-30m&latest=now&sid=1741879088.60544). Would you want some mechanism by which RTR informs Glides that the vehicle is being deleted, or is that being handled elsewhere?


For completion, this isn't actually how it works for LRVs. These are cleaned up by the `remove_stale_vehicles` function which deletes any vehicles that haven't received any updates for the past 30 minutes. You can search for logs with `deleted_stale_vehicle` from RTR.

rudiejd: @mathcolo  there is no context attribute in the CloudEvents schema for denoting a deletion, so you would have to shoehorn it into data.

IME the deletion of a train is separate from this information event. how about a separate event for saying that a vehicle has been deleted?

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
```diff
@@ -0,0 +1,39 @@
...
+- `stopStatus` (enum, values: ["STOPPED_AT", "INCOMING_AT", "IN_TRANSIT_TO"]): What the train is doing in relation to the next stop
+- `lat` (string): The latitude of the train, provided by the AVLThe latitude of the train, provided by the AVLThe latitude of the train, provided by the AVL
+- `lon` (string): The longitude of the train, provided by the AVL
```
lemald: **nitpick:** what's "AVL" in this context - just automatic vehicle location in general, or was it a typo of "AVI?" My $0.02 is that the documentation doesn't need to get into too much depth about the source of the data here other than that it's the most up-to-date location data that RTR has available.

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
lemald: Maybe the documentation could explicitly say that it's going to be a specific platform stop ID.

schemas/com.mbta.rtr.lightrail-vehicle-information.json
```diff
@@ -0,0 +1,92 @@
...
+          },
+          "examples": [["3644", "3862"]]
+        },
+        "orientation": {
+          "type": "string",
+          "enum": ["AB", "BA"]
+        },
+        "nextStopId": {
+          "type": "string",
+          "examples": ["place-gover", "place-clmnl", "place-pktrm"]
```
lemald: **nitpick:** since these are parent stop IDs rather than individual platforms, these aren't actual examples we would see in the feed.

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
rudiejd: that makes sense to me. removed the bit about AVL

examples/com.mbta.rtr.lightrail-vehicle-information/LightRailVehicleInformation.json
```diff
@@ -0,0 +1,20 @@
...
+      "stopStatus": "IN_TRANSIT_TO",
+      "lat": "42.21469",
+      "lon": "71.03305",
+      "heading": "26.69"
+  },
+  "id": "1",
+  "source": "ctd.mbta.com/rtr-1234",
+  "specversion": "1.0",
+  "time": "2025-03-12T00:01:00.110000Z",
+  "type": "com.mbta.rtr.lightrail-vehicle-information"
```
rudiejd: what do you all think of the naming here? Another idea was `lightrail-vehicle-heartbeat` to maintain parity with the subwayside vehicle heartbeat

docs/events/com.mbta.rtr.lightrail-vehicle-information.mdx
rudiejd: I added a separate event for deletion that only specifies `vehicleId`. let me know what you think

schemas/com.mbta.rtr.lightrail-vehicle-information.json
```diff
@@ -0,0 +1,92 @@
...
+          },
+          "examples": [["3644", "3862"]]
+        },
+        "orientation": {
+          "type": "string",
+          "enum": ["AB", "BA"]
+        },
+        "nextStopId": {
+          "type": "string",
+          "examples": ["70160", "70161", "place-pktrm"]
```
lemald: **nitpick:** you still list `place-pktrm` here, which is a parent stop rather than a platform.

rudiejd: oops

examples/rtr-events/LightRailVehicleInformation.json
```diff
@@ -0,0 +1,20 @@
...
+      "lastAvi": "875",
+      "vehicleId": "G-10023",
+      "routeId": "Green-D",
+      "consist": ["3644", "3862"],
+      "orientation": "AB",
+      "nextStopId": "70203",
+      "stopStatus": "IN_TRANSIT_TO",
+      "lat": "42.21469",
+      "lon": "71.03305",
+      "heading": "26.69"
```
mathcolo: Should heading be removed from the example, as it's not in the schema?

examples/rtr-events/LightRailVehicleInformation.json
```diff
@@ -0,0 +1,20 @@
...
+      "directionId": 1,
+      "lastAvi": "875",
+      "vehicleId": "G-10023",
+      "routeId": "Green-D",
+      "consist": ["3644", "3862"],
+      "orientation": "AB",
+      "nextStopId": "70203",
+      "stopStatus": "IN_TRANSIT_TO",
+      "lat": "42.21469",
+      "lon": "71.03305",
```
mathcolo: This will hopefully be `-71`, if it's in decimal degrees?

docs/events/rtr/com.mbta.rtr.lightrail-vehicle-deletion.mdx
```diff
@@ -0,0 +1,31 @@
...
+
+RTR deletes a light rail vehicle when 30 minutes have elapsed without any messages pertaining to said vehicle.
+
+Related to [RFC-25](https://github.com/mbta/technology-docs/blob/main/rfcs/accepted/0025-light-rail-trip-model.md).
+
+## Data
+
+Fields in the event's `data`:
+
+- `vehicleId` (string): Unique identifier of the vehicle from OCS. Same `vehicleId` as [com.mbta.rtr.lightrail-vehicle-information](com.mbta.rtr.lightrail-vehicle-information#Data).
```
mathcolo: Nice, I like this better! will deletions be issued all remaining vehicles at the end of the service day, or do consumers need to do our own pruning as well?

examples/rtr-events/LightRailVehicleInformation.json
rudiejd: good call, yes that is unnecessary

examples/rtr-events/LightRailVehicleInformation.json
rudiejd: yep sorry, that should be -71

docs/events/rtr/com.mbta.rtr.lightrail-vehicle-deletion.mdx
rudiejd: i am probably missing somethiong - could you explain why this would be necessary? My assumption would be that during the non-revenue period either a vehicle will go out of service, in which case 10 minutes will elapse without any message and the vehicle will be deleted, or it will continue making some kind of ad-hoc trips, in which case the vehicle will not be deleted because it will still be doing trips, even if they are non-revenue. I'm not sure why there would need to be specific logic for deleting a vehicle at the end of a service day, and RTR currently doesn't have this logic.

mathcolo: Ah, so I was only wondering because of the rare possibility of deletion messages not making it over for some reason. i.e. if a deletion doesn't make it over, will it get "stuck" in Glides?

examples/com.mbta.rtr.lightrail-vehicle-information/LightRailVehicleInformation.json
mathcolo: Hmm, what subwayside is describing seems quite distinct. what about `lightrail-movement`?

docs/events/rtr/com.mbta.rtr.lightrail-vehicle-deletion.mdx
rudiejd: yeah, in a kinesis outage there would be vehicles in glides that do not exist in RTR. but I can't see any way to avoid the issue unless you want to add a fallback form of communication e.g. falling back to a synchronous HTTP call. But I think I'd probably be over-engineering if I start thinking of a fallback mechanism before we've had issues with the first communication channel. My proposal would be to start with the deletion event that we have here, then iterate (add and end-of-day deletion mechanism in RTR, add pruning in glides, etc.) if it proves to be insufficient

examples/com.mbta.rtr.lightrail-vehicle-information/LightRailVehicleInformation.json
rudiejd: i feel like `lightrail-movement` implies that these will only be sent when the vehicle moves. wouldn't you want to see  periodic updates even if the vehicle is not moving?

mathcolo: Ah, yes. hmm. what about `update`? There's nothing specifically wrong with `information`, it's just slightly generic (and I don't want to miss an opportunity to make sure it's clear).

docs/events/rtr/com.mbta.rtr.lightrail-vehicle-deletion.mdx
mathcolo: alright, that works for me 👍

examples/com.mbta.rtr.lightrail-vehicle-information/LightRailVehicleInformation.json
rudiejd: Update works for me! I was searching for something better than information - generally I prefer events to have action verbs

rudiejd: I renamed this event to `lightrail-vehicle-updated` and I renamed the other event to `lightrail-vehicle-deleted`. Using (object)-(past tense verb) makes these events match up nicely with the existing glides events in the schemas repo.

docs/events/rtr/com.mbta.rtr.lightrail-vehicle-information.mdx
```diff
@@ -0,0 +1,42 @@
+---
+sidebar_label: rtr.lightrail-vehicle-updated
+---
+
+import LightRailVehicleInformation from "@site/examples/rtr-events/LightRailVehicleInformation.json";
+import EventSchemaPath from "@site/src/components/EventSchemaPath.tsx";
+import Examples from "@site/src/components/Examples.tsx";
+
+# lightrail-vehicle-updated
```
mathcolo: sorry to be pedantic, perhaps `update` instead of `updated`? i.e. noun instead of past participle.

schemas/com.mbta.rtr.lightrail-vehicle-updated.json
```diff
@@ -0,0 +1,92 @@
...
+    "source": {
+      "$ref": "cloudevents.json#/properties/source",
+      "examples": ["mbta.com/rtr-1"]
+    },
+    "specversion": {
+      "const": "1.0",
+      "$ref": "cloudevents.json#/properties/specversion"
+    },
+    "type": {
+      "const": "com.mbta.rtr.lightrail-vehicle-update",
```
mathcolo: this is already `update`?

docs/events/rtr/com.mbta.rtr.lightrail-vehicle-information.mdx
rudiejd: I'm using past participle for consistency with the glides events: https://github.com/mbta/schemas/blob/main/docs/events/glides/com.mbta.ctd.glides.editors_changed.v1.mdx

I think it should either be all one tense or all the other, and it wouldn't be very convenient to change the existing events to present tense. Plus if you want to really split hairs, by the time you receive an event it has already "happened" on the producer's side so past participle could be more appropriate

mathcolo: 👍 gotcha! makes sense, thanks for pointing that out

schemas/com.mbta.rtr.lightrail-vehicle-updated.json
rudiejd: oops, yeah this was a mistake

mathcolo (approved):
Thank you for adding heading back!

rudiejd:
Problem:
Glides needs information about light rail vehicles to start doing trip assignment on its own. Further outlined here: https://www.notion.so/mbta-downtown-crossing/Glides-Trips-RTR-Inter-dependency-Problem-Statement-185f5d8d11ea80d5a7b5e8c24cb85ce7

Solution:
Add com.mbta.rtr.lightrail-vehicle-information message type with all of the necessary information for Glides LRV trip assignment.

PR #13, opened 2025-04-25
feat(lightrail-vehicle-updated): add timestamp for vehicle to light rail updated event

rudiejd:
Ticket: ad-hoc

Problem:
The CloudEvents specification provides a `time` field, which represents the ["timestamp of when the occurrence [underlying the event] happened"](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#time). The "occurrence" in this case is an update to the `Vehicle` state in RTR. However, we would also like to know the last time that OCS sent a TMOV for a vehicle, which might be different from the last time that RTR updated the vehicle.

Solution:
Send an `updatedAt` field, analogous to `updated_at` in the `/vehicles` [API endpoint](https://api-v3.mbta.com/vehicles) or `updatedAt` in GTFS-RT. This signifies the last time RTR received an update from OCS regarding the vehicle.


lemald:
A couple of initial thoughts:

1. Could you also add a Glides developer for review?
2. Since this is adding a new required field this is technically a breaking change; do we feel okay doing this without a version bump and everything given that there are no consumers fully implemented yet? I'm leaning towards "yes," but I want to make sure it's a conscious decision.

rudiejd:
> Since this is adding a new required field this is technically a breaking change; do we feel okay doing this without a version bump and everything given that there are no consumers fully implemented yet? I'm leaning towards "yes," but I want to make sure it's a conscious decision.

yes, I feel like this is fine without a version bump since the only consumer is glides and this consumer is not yet live.

mathcolo (commented):
Would this be better as `timestamp` to be closer to [GTFS-RT](https://gtfs.org/documentation/realtime/reference/#message-feedheader)?

rudiejd:
@mathcolo we could do that! my thinking was that it might be a little confusing since there is also a timestamp for the GTFS-RT feed, and there's a timestamp for the CloudEvent (`time`), so I felt like `updatedAt` was more descriptive

lemald (approved):

PR #15, opened 2025-05-22
feat(rtr): document that consist can be null in RTR vehicles feed

lemald:
Per conversation between myself, @mathcolo, and @rudiejd, updating the spec to reflect the fact that a consist can be `null` (as opposed to an empty array) if we don't have that data for a vehicle for whatever reason.


docs/events/rtr/com.mbta.rtr.lightrail-vehicle-updated.mdx
rudiejd, line 1: could we say why the consist is null if/when that is the case?

lemald: Hmm... is there anything we could really say that's meaningfully more descriptive than "it's`null` when we don't have that information?"

rudiejd: if we don't have any example situations in which the consist can be null, then I think it's fine to just keep the type

rudiejd (approved):

mathcolo (approved):

PR #17, opened 2026-02-26
feat(glides): add revenue field to trip assignment message

rudiejd:
Asana Ticket: [🔮🐛 Investigate and fix problematic revenue predictions](https://app.asana.com/1/15492006741476/project/584764604969369/task/1212917030453989?focus=true)

Adds a revenue field to the `VehicleTripAssignment` message from Glides. See https://github.com/mbta/glides/pull/3251 for a description of why this field was necessary.


lemald (approved):
I don't _love_ adding more trip metadata to the vehicle trip assignment messages, but I agree that it's probably the best way to solve the issue at hand for now.

rudiejd:
@lemald yeah, this is a "stop the bleeding" type of change in my view as well. Long term, we should implement TODS and share the same concept of revenue and non-revenue trips department wide. If that ever happens, we can trim this field from the trip assignment message.

PR #20, opened 2026-06-11
feat(glides): add `nextTripKey` field to trips_updated.v1

andrewdolce:
Pullout inspectors using Glides may decide to assign a vehicle to an upcoming trip in a way that breaks the vehicle's expected block of trips from the HASTUS schedule. Currently Glides can only convey this information via a `vehicle_trip_assignment.v1` message, which cannot be emitted until the vehicle has finished turning around and is in position to begin the trip. Unfortunately this may be too late for RTR to make timely terminal predictions.

The solution is to allow Glides to indicate the change in the block sequence ahead of time as part of a `trips_updated.v1` message, by populating a `nextTripKey` field. (This is somewhat analogous to the `TSCH LNK` messages used by OCS to indicate block changes for heavy rail.)

### DIFF
--- a/docs/events/glides/com.mbta.ctd.glides.trips_updated.v1.mdx
+++ b/docs/events/glides/com.mbta.ctd.glides.trips_updated.v1.mdx
@@ -108,7 +108,7 @@ New restrictions on existing fields:
 - At least one of `startTime`, `endTime`, and `previousTripKey` is required.
 - `revenue` (`"revenue"` | `"nonrevenue"`, optional): If this field is missing when adding a trip, the default is that the trip is in revenue service.
 - `dropped` SHOULD NOT be set.
-- `nextTripKey` SHOULD NOT be set.
+- `nextTripKey` SHOULD NOT be set. By convention, the `previousTripKey` field should be used instead to indicate any sequence between added trips.
 - All data SHOULD NOT be `"none"` or `"unset"`. (Instead, the fields would not be included).
 
 Glides SHOULD include the `cars` field to indicate the length of the train, even if no information is known about each car.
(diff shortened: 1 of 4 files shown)


docs/events/glides/com.mbta.ctd.glides.trips_updated.v1.mdx
```diff
@@ -179,6 +180,7 @@ Fields in the object:
...
+- `nextTripKey` ([TripKey](#tripkey), optional): If present, this specifies the key of another trip that will happen immediately after this trip. This field MAY refer to a trip that has not been already seen in the event stream. This field is intended to communicate a change in the expected sequence of trips, such as the sequence implied by blocks in the schedule from HASTUS. Note that a single change indicating a sequential link between two trips may imply additional changes to other links. For example, consider an initial schedule containing sequences between trips A1 --> A2, and B1 --> B2. If Glides emits an update to A1 with `nextTripKey` referencing B2, this implies a break to the two initial links. Clients are expected to maintain a consistent schedule such that no two trips may link to the same "next" trip. Note also that this field does not support an "unset" value. If Glides wishes to restore the initial schedule after changing some links, it must explicitly emit the links to be restored.
```
lemald: **suggestion:** a couple of things:

1. Reword this:

> If present, this specifies the key of another trip that will happen immediately after this trip.

to something like

> If present, this specifies the key of another trip that the vehicle performing this trip will perform immediately after.

2. Re:

> this implies a break to the two initial links.

can you be explicit and list out what links would be broken and which would be formed by processing the message?

andrewdolce: I'll take another pass at expanding it. Might break this out into a little subsection since it's getting a bit long to be an in-line property description?

andrewdolce:
I took the rewording suggestion and expanded the block break example. I also added a callout for the lack of a `"none"` value, which I'm still slightly up in the air about? I don't yet see a reason we'd need it, but had considered adding it just in case it becomes useful in the future. Going to leave as-is, but open to feedback on that.

lemald (approved):
Looks good to me! Not sure if @rudiejd has any thoughts to add.

rudiejd:
this looks good to me! sorry, I reviewed but forgot to leave an approval

rudiejd (approved):

rudiejd:
one note is that I think it might simplify this schema for glides to send `trip_id` as the identifier instead of `trip_key`. RTR uses the `trip_id` for deduplication internally, so right now the information in the `trip_key` is simply redundant given the rest of the message. but I realize that does not match with our current data model

andrewdolce:
> one note is that I think it might simplify this schema for glides to send `trip_id` as the identifier instead of `trip_key`. RTR uses the `trip_id` for deduplication internally, so right now the information in the `trip_key` is simply redundant given the rest of the message. but I realize that does not match with our current data model

How strongly do you feel about this? Even if it's redundant, my instinct is to stick with trip key to keep it consistent with the other fields. Also just to clarify, you're talking about for the next trip right? Not changing the existing `tripKey` field on the update message (which I feel like would be a non-starter)?

andrewdolce:
(@rudiejd Forgot to tag you in my reply)

rudiejd:
My comment was not blocking, but for reference I think all of the references to `trip_key` should use `trip_id` instead.

